PR #1, opened 2020-01-08
SYN-36: Migrating SDDs to Github

akosma:
This PR contains the infrastructure required for the migration of SDDs from the wiki to Github:
- Navigation file
- Template file
- Meta info table template partial


akosma:
This is a WIP for the moment.

akosma:
Ready for review

docs/modules/SDDs/nav.adoc
```diff
@@ -0,0 +1,22 @@
+.Syn Design Documents - SDD
```
tobru, line 1: Should we maybe update the navigation per added SDD? We will most likely not add every SDDs at once which will eventually lead into broken navigation...

akosma: That will lead to conflicts I guess, because all SDD branch out from `master` and would all modify the same empty file. I thought since we know what the filenames will be, just provide the navigation and it will eventually be fully working.

tobru: OK, let's try. We can change it later anytime =)

tobru (approved):

PR #5, opened 2020-01-13
SYN-36: SDD 0004 - Application deployments

akosma:
Contents of SDD 0004


tobru:
This is IMHO not ready to be published

tobru:
This SDD will be replaced by a new one and I therefore close this one.

PR #6, opened 2020-01-13
SYN-36: SDD 0005 - Crossplane Service Provisioning

akosma:
Contents of SDD 0005


tobru:
Needs some more generalization, more being Project Syn and less being VSHN Syn.

tobru:
This SDD will be replaced by a new one.

PR #8, opened 2020-01-13
SYN-36: SDD 0007 - Lieutenant - Management API

akosma:
Contents of SDD 0007


tobru:
Probably needs some re-copy-paste-ing of content from Confluence. Will do that in a follow-up PR.

PR #9, opened 2020-01-13
SYN-36: SDD 0008 - Platform Configuration Management

akosma:
Contents of SDD 0008


tobru:
Will need some more generalization for Project Syn, less VSHN stuff...

srueg (approved):

PR #11, opened 2020-01-13
SYN-36: SDD 0010 - Syn Overall Concept

akosma:
Contents of SDD 0010


tobru:
This will need some more de-VSHNization...

tobru:
This SDD contains too much VSHN stuff and doesn't add any value to Project Syn. Therefore I close this one.

PR #17, opened 2020-01-13
SYN-36: SDD 0016 - Backup

akosma:
Contents of SDD 0016


tobru:
Needs some more de-VSHNization

tobru:
This SDD doesn't add any value to Project Syn.

PR #22, opened 2020-01-24
Various changes

Kidswiss:
To reflect what's actually implemented.


srueg (approved):

srueg:
@tobru FYI we removed the `slug`. If we really need them at some point we'll add them again.

PR #47, opened 2020-06-10
SDD 0020 - OpenShift Hive

srueg:
Write first draft of SDD 0020 - OpenShift Hive

### DIFF
--- a/docs/modules/SDDs/pages/0020-openshift-hive.adoc
+++ b/docs/modules/SDDs/pages/0020-openshift-hive.adoc
@@ -84,6 +84,20 @@ For provisioning we use a default setup of three master and three worker nodes.
 Once a cluster is provisioned, scaling will be implemented via other means for example in a Commodore component.
 
 
+==== Cluster Synfection
+
+Once a cluster is provisioned via Hive the next step is to synfect it (install Steward).
+Hive provides the `SyncSet` CRD to create and patch arbitrary resources on a provisioned cluster.
+While this mechanism could be used to install the Steward agent on a new cluster it also poses some downsides:
+
+* We would need to duplicate the installation manifests (currently implemented in the xref:0007-lieutenant-management-api.adoc[Lieutenant API]) to create a `SyncSet` out of them
+* Once Syn is fully bootstrapped on the cluster, Steward itself will be managed by xref:0019-gitops.adoc[GitOps]. This would end up with two systems managing the same resources (Hive and GitOps)
+
+Instead of using a `SyncSet` we use the credentials of a provisioned cluster and run a job which installs Syn.
+This can be implemented relatively easy since the credentials (Kubeconfig) for a cluster are stored in a secret.
+The controller can create the Kubernetes job which mounts this secret and runs `kubectl apply -f` against the install URL of the respective cluster.
+
+
 === Risks and Mitigations
 
 With this design we're relatively tightly coupled to the Hive operator as in the created CRs (API) are defined by it.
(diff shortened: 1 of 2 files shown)


docs/modules/SDDs/pages/0020-openshift-hive.adoc
```diff
@@ -0,0 +1,101 @@
...
+==== Provisioning Information
+
+In order for this controller to be able to create the necessary objects, it needs to receive certain information.
+All confidential information (like cloud provider credentials or image pull secrets) should be stored in Kubernetes secrets and referenced.
+In a first PoC phase, all information that's not yet present in a cluster's facts will be provided in annotations on the cluster object:
+
+* `hive.syn.tools/gcp-project*` GCP project name
+* `hive.syn.tools/base-domain*` Base domain
+* `hive.syn.tools/credentials-secret*` Name of the secret containing the cloud credentials
+* `hive.syn.tools/worker-count` Count of worker nodes to provision
```
corvus-ch: I doubt that scaling is something we should do via Hive. This is something we can do in a more robust way using a component.

From my point of view, it would be find to just go with the defaults for bootstrapping the cluster and then let the component to take over.

So drop `worker-count` and `worker-type`.

docs/modules/SDDs/pages/0020-openshift-hive.adoc
```diff
@@ -0,0 +1,101 @@
...
+For example if a cluster object has the annotation `syn.tools/cluster-provisioner=hive` set, an OpenShift cluster should be provisioned via Hive.
+
+
+==== Provisioning Information
+
+In order for this controller to be able to create the necessary objects, it needs to receive certain information.
+All confidential information (like cloud provider credentials or image pull secrets) should be stored in Kubernetes secrets and referenced.
+In a first PoC phase, all information that's not yet present in a cluster's facts will be provided in annotations on the cluster object:
+
+* `hive.syn.tools/gcp-project*` GCP project name
```
corvus-ch: Use a footnote instead of a plain `*`.

corvus-ch (changes requested):

tobru:
I think that the ideas of using annotations first is a good thing! Changing this into a proper struct of the CRD later makes it much easier to have a PoC ready.

docs/modules/SDDs/pages/0020-openshift-hive.adoc
```diff
@@ -84,6 +84,20 @@ For provisioning we use a default setup of three master and three worker nodes.
 Once a cluster is provisioned, scaling will be implemented via other means for example in a Commodore component.
 
 
+==== Cluster Synfection
+
+Once a cluster is provisioned via Hive the next step is to synfect it (install Steward).
+Hive provides the `SyncSet` CRD to create and patch arbitrary resources on a provisioned cluster.
+While this mechanism could be used to install the Steward agent on a new cluster it also poses some downsides:
+
+. We would need to duplicate the installation manifests (currently implemented in the xref:0007-lieutenant-management-api.adoc[Lieutenant API]) to create a `SyncSet` out of them
```
corvus-ch: Is the order relevant? I doubt and would use an unordered list here.

docs/modules/SDDs/pages/0020-openshift-hive.adoc
```diff
@@ -84,6 +84,20 @@ For provisioning we use a default setup of three master and three worker nodes.
...
+. Once Syn is fully bootstrapped on the cluster, Steward itself will be managed by xref:0019-gitops.adoc[GitOps]. This would end up with two systems managing the same resources (Hive and GitOps)
+
+An alternative to using a `SyncSet` would be to use the credentials of a provisioned cluster and run a job which installs Syn.
```
corvus-ch: I would not name this an _alternative_ but the suggested/chosen approach.

corvus-ch (changes requested):

corvus-ch (approved):

PR #48, opened 2020-06-10
SDD 0021 - Cluster Compilation

srueg:
Initial draft for SDD 0021 - Cluster Compilation

Input from https://wiki.vshn.net/display/SYN/2020-06-10+Meeting+notes+Commodore+Job+Runner

### DIFF
--- a/docs/modules/SDDs/pages/0021-cluster-catalog-compilation.adoc
+++ b/docs/modules/SDDs/pages/0021-cluster-catalog-compilation.adoc
@@ -51,7 +51,7 @@ By using the Kapitan inventory the tool needs to figure out for which clusters t
 In a first step this could be relatively rudimentary and only use the input information from a webhook (cluster/tenant ID) to decide which clusters to compile.
 
 The information if a cluster needs to be compiled is then stored in the cluster object of the Lieutenant API.
-In a first PoC phase this can be an annotation like `cluster.syn.tools/needs-compilation=true`.
+In a first PoC phase this can be an annotation like `cluster.syn.tools/compilation-id=DmpKUlLdCfUD`.
 In a later stage this information should be part of the typed struct and might include further information like the exact version of the various input repos.
 
 
--- a/docs/modules/SDDs/pages/0021-cluster-catalog-compilation.adoc
+++ b/docs/modules/SDDs/pages/0021-cluster-catalog-compilation.adoc
@@ -1,0 +1,132 @@
+= SDD 0021 - Cluster Catalog Compilation
+
+:sdd_author:    Simon Rüegg
+:sdd_owner:
+:sdd_reviewers:
+:sdd_date:      2020-06-10
+:sdd_status:    draft
+include::partial$meta-info-table.adoc[]
+
+[NOTE]
+.Summary
+====
+This document describes how, where and when the xref:0006-commodore-configuration-catalog-compilation.adoc[catalog compilation engine] runs.
+====
+
+
+== Motivation
+
+The compilation of a cluster catalog is a central piece of the configuration management within project Syn.
+It generates the actual configuration for a cluster based on various inputs and configurations.
+It's therefore crucial to have a robust and scalable system in place to run these compilation jobs in a secure and reactive manner.
+
+
+=== Goals
+
+* Define where (in terms of which cluster, namespace, etc.) the jobs are run
+* Define what should cause the recompilation of the catalog for a certain cluster
+* Scale to several hundrets and with future changes to thousands of clusters
+
+
+=== Non-Goals
+
+* The handling of a compiled catalog (for example merge request automation) is out of scope for this design document
+* Solution to automate the component updates
+
+
+== Design Proposal
+
+The design splits the problem into two parts:
+
+. Figuring out which cluster catalogs need to be compiled
+. Running the actual compilation job
+
+
+=== Quartermaster
+
+A new tool called https://en.wikipedia.org/wiki/Quartermaster#Royal_Navy[Quartermaster] is implemented.
+This tool should solve the problem of figuring out which cluster catalogs need to be recompiled.
+The tool gets input in the form of webhooks which are triggered by Git repositories or other tools.
+By using the Kapitan inventory the tool needs to figure out for which clusters the config changed and therefore needs a recompilation.
+In a first step this could be relatively rudimentary and only use the input information from a webhook (cluster/tenant ID) to decide which clusters to compile.
+
+The information if a cluster needs to be compiled is then stored in the cluster object of the Lieutenant API.
+In a first PoC phase this can be an annotation like `cluster.syn.tools/needs-compilation=true`.
+In a later stage this information should be part of the typed struct and might include further information like the exact version of the various input repos.
+
+
+==== Webhooks
+
+Triggering a webhook can be implemented in either a Git repo natively (for example https://developer.github.com/webhooks/[GitHub Webhooks]) or via a CI/CD system (for example a https://docs.gitlab.com/ee/ci/[GitLab CI] job which runs `curl`).
+A webhook can also be called by any other tool or even manually with a `curl` command.
+
+A webhook can include the following optional information:
+
+* Cluster ID for which to compile the catalog.
+This can be used in case where a specific cluster should be compiled
+
+* Tenant ID for which to figure out affected clusters.
+This can be used to limit the set of clusters to be considered
+
+* Set of classes that changed
+
+
+==== Authentication & Authorization
+
+To prevent exsessive resource usage or even DoS attacks, the webhooks need to be authenticated.
+A permission system needs to be in place to decide which entities are allowed to trigger which actions.
+It shouldn't be possible for example for a tenant to trigger compilation for clusters of other tenants.
+
+
+=== User Stories
+
+==== Config Change in Global Config Repo
+
+If configuration in the global inventory repo changes, Quartermaster needs to find all clusters which are affected by the changed configuration.
+
+
+==== Config Change in Tenant Config Repo
+
+If configuration in a tenant inventory repo changes, Quartermaster needs to find all clusters which are affected by the changed configuration.
+The set of clusters to consider are the clusters of the respective tenant.
+
+
+==== Change of Cluster Facts
+
+If the facts of a cluster change the set of clusters to consider already only consists of this cluster.
+Triggering the webhook needs to be implemented in Lieutenant (operator or API).
+
+
+==== New Release of a Commodore Component
+
+The new release of a Commodore component won't trigger anything directly.
+If the specific version for a component is being updated it results in a config change in either the global inventory repo or a tenant repo.
+This in turn results in the recompilation of the affected clusters.
+
+
+==== Manually Trigger a Compile
+
+To manually trigger a compile for developping or debugging purposes the webhook can be called with a tool like `curl`.
+Providing a cluster or tenant ID can limit the set of clusters to compile.
+
+
+=== Risks and Mitigations
+
+==== Rate Limiting
+
+Since the compilation is a relatively expensive operation resource wise, it might become necessary to implement some kind of rate limiting.
+For example could Quartermaster make sure that a certain amount of compile jobs per amount of time isn't exceeded for a single cluster or tenant.
+This would mitigate the situation where a lot of changes in various Git repos would result in many compile jobs which could exhaust the resources of a cluster.
+The downside would be a longer reaction time for changes to be applied on a cluster.
+
+
+== Alternatives
+
+An alternative would be to couple the two mentioned parts of the design together.
+This would mean that the same component is responsible to trigger new compiles and run the compiles.
+Such a solution would probably require a centralized
+
+
+== References
+
+* xref:0006-commodore-configuration-catalog-compilation.adoc[Configuration catalog compilation]
--- a/docs/modules/SDDs/pages/0021-cluster-catalog-compilation.adoc
+++ b/docs/modules/SDDs/pages/0021-cluster-catalog-compilation.adoc
@@ -23,7 +23,7 @@ It's therefore crucial to have a robust and scalable system in place to run thes
 
 === Goals
 
-* Define where (in terms of which cluster, namespace, etc.) the jobs are run
+* Define where (in terms of which cluster, namespace, etc.) the jobs run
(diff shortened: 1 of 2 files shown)


docs/modules/SDDs/pages/0021-cluster-catalog-compilation.adoc
```diff
@@ -0,0 +1,145 @@
...
+For example could Quartermaster make sure that a certain amount of compile jobs per amount of time isn't exceeded for a single cluster or tenant.
+This would mitigate the situation where a lot of changes in various Git repos would result in many compile jobs which could exhaust the resources of a cluster.
+The downside would be a longer reaction time for changes to be applied on a cluster.
+
+
+== Alternatives
+
+An alternative would be to couple the two mentioned parts of the design together.
+This would mean that the same component is responsible to trigger new compiles and run the compiles.
+Such a solution would probably require a centralized
```
simu: this sentence looks incomplete, probably should be "require a centralized job runner" or something along those lines?

docs/modules/SDDs/pages/0021-cluster-catalog-compilation.adoc
```diff
@@ -0,0 +1,145 @@
...
+
+
+=== Compilation Jobs
+
+To compile the catalog for a cluster, a compilation job is created.
+The compilation job is a https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion[Kubernetes job] resource which executes the xref:0006-commodore-configuration-catalog-compilation.adoc[catalog compilation engine].
+
+To distribute the load this compilation generates and to follow a decentralized approach, these jobs are run on the same cluster as they compile the catalog for.
+The Steward component on a cluster starts a compilation job if the `cluster.syn.tools/needs-compilation` annotation is set to `true` on its own cluster object.
+It will then launch the respective job to compile the cluster catalog.
```
simu, line 102: I would like to see at least a sketch of how Steward will launch the job to compile the catalog in this SDD.

Right now we're using argo-events to create plain Kubernetes jobs to compile catalogs. I don't think it's necessary to have argo-events on each cluster, but we probably want to have some compile job template that can be instantiated by Steward when a catalog needs to be compiled.

docs/modules/SDDs/pages/0021-cluster-catalog-compilation.adoc
```diff
@@ -0,0 +1,145 @@
...
+
+For clusters where this isn't possible due to resource, security or other constraints, a central system runs the jobs.
+These clusters need to be configured in the central system so it will react on the mentioned annotation an create the compile jobs for them.
+
+
+=== User Stories
+
+==== Config Change in Global Config Repo
+
+If configuration in the global inventory repo changes, Quartermaster needs to find all clusters which are affected by the changed configuration.
```
simu, line 121: Probably a good idea to discuss the problems we're facing here, and sketching a first approach for actually determining which clusters are affected.

Having thought about this a bit more, I think we can get away with completely ignoring component classes, and just backtrace whether changes to a particular class affect a given cluster by constructing the cluster target and mocking all component classes as empty classes, and then listing all included classes using reclass.

docs/modules/SDDs/pages/0021-cluster-catalog-compilation.adoc
```diff
@@ -0,0 +1,145 @@
...
+
+
+==== Config Change in Tenant Config Repo
+
+If configuration in a tenant inventory repo changes, Quartermaster needs to find all clusters which are affected by the changed configuration.
```
simu, line 126: Same mechanism as for changes in the global repo can be used here, for a reduced set of clusters.

simu (commented):
LGTM so far.

I've added a few comments for things that I think are missing, but should be in the SDD

docs/modules/SDDs/pages/0021-cluster-catalog-compilation.adoc
srueg: I was thinking that Steward would just create the Job via the K8s API (so the "template" would be in go code)

docs/modules/SDDs/pages/0021-cluster-catalog-compilation.adoc
srueg: explicitly mentioned this

docs/modules/SDDs/pages/0021-cluster-catalog-compilation.adoc
tobru: How do we find out when the compile job fails? Maybe we need some Prometheus rules explicitly for that.?

srueg: Good point.
We could push metrics to the Prom pushgateway from the compile job itself.

srueg: I added a paragraph for each inventory parsing and metrics respectively.

docs/modules/SDDs/pages/0021-cluster-catalog-compilation.adoc
```diff
@@ -0,0 +1,170 @@
...
+== Motivation
+
+The compilation of a cluster catalog is a central piece of the configuration management within project Syn.
+It generates the actual configuration for a cluster based on various inputs and configurations.
+It's therefore crucial to have a robust and scalable system in place to run these compilation jobs in a secure and reactive manner.
+
+
+=== Goals
+
+* Define where (in terms of which cluster, namespace, etc.) the jobs are run
```
corvus-ch:
```suggestion
* Define where (in terms of which cluster, namespace, etc.) the jobs run
```

docs/modules/SDDs/pages/0021-cluster-catalog-compilation.adoc
```diff
@@ -0,0 +1,170 @@
...
+
+* The handling of a compiled catalog (for example merge request automation) is out of scope for this design document
+* Solution to automate the component updates
+
+
+== Design Proposal
+
+The design splits the problem into two parts:
+
+. Figuring out which cluster catalogs need to be compiled
```
corvus-ch, line 41: Is the order relevant? If not, make this an unordered list.

simu (approved):

docs/modules/SDDs/pages/0021-cluster-catalog-compilation.adoc
```diff
@@ -0,0 +1,170 @@
...
+=== Quartermaster
+
+A new tool called https://en.wikipedia.org/wiki/Quartermaster#Royal_Navy[Quartermaster] is implemented.
+This tool should solve the problem of figuring out which cluster catalogs need to be recompiled.
+The tool gets input in the form of webhooks which are triggered by Git repositories or other tools.
+By using the Kapitan inventory the tool needs to figure out for which clusters the config changed and therefore needs a recompilation.
+In a first step this could be relatively rudimentary and only use the input information from a webhook (cluster/tenant ID) to decide which clusters to compile.
+
+The information if a cluster needs to be compiled is then stored in the cluster object of the Lieutenant API.
+In a first PoC phase this can be an annotation like `cluster.syn.tools/needs-compilation=true`.
```
corvus-ch: Maybe we should store a timestamp here. By doing so, the executing part does not need to update the object.

* Quartermaster figures a rebuild is needed and writes the current timestamp into the annotation.
* The thing responsible for trigger the build jobs, compares the timestamp in the annotation with the timestamp of the last execution. If the last execution is older than what is stored in the annotation, a new build gets triggered.

docs/modules/SDDs/pages/0021-cluster-catalog-compilation.adoc
```diff
@@ -0,0 +1,170 @@
...
+
+* Set of classes that changed
+
+
+==== Authentication & Authorization
+
+To prevent exsessive resource usage or even DoS attacks, the webhooks need to be authenticated.
+A permission system needs to be in place to decide which entities are allowed to trigger which actions.
+It shouldn't be possible for example for a tenant to trigger compilation for clusters of other tenants.
+
```
corvus-ch, line 93: Since we are talking about webhooks, this kind of authentication must work as part of a GET request and thus has to be part of the URL.

docs/modules/SDDs/pages/0021-cluster-catalog-compilation.adoc
```diff
@@ -0,0 +1,170 @@
...
+To compile the catalog for a cluster, a compilation job is created.
+The compilation job is a https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion[Kubernetes job] resource which executes the xref:0006-commodore-configuration-catalog-compilation.adoc[catalog compilation engine].
+
+To distribute the load this compilation generates and to follow a decentralized approach, these jobs are run on the same cluster as they compile the catalog for.
+The Steward component on a cluster starts a compilation job if the `cluster.syn.tools/needs-compilation` annotation is set to `true` on its own cluster object.
+It will then launch the respective job to compile the cluster catalog.
+
+For clusters where this isn't possible due to resource, security or other constraints, a central system runs the jobs.
+These clusters need to be configured in the central system so it will react on the mentioned annotation an create the compile jobs for them.
```
corvus-ch, line 105: So will it really be Steward creating the job objects? At least the code we write to do so needs to be reusable so it can be integrated in Steward but also be used on surrogate systems hosting the compilation for another cluster.

I imagine this to be an project on APPUiO per tenant with surrogate catalogue compilation needs.

docs/modules/SDDs/pages/0021-cluster-catalog-compilation.adoc
```diff
@@ -0,0 +1,170 @@
...
+
+==== Change of Cluster Facts
+
+If the facts of a cluster change the set of clusters to consider already only consists of this cluster.
+Triggering the webhook needs to be implemented in Lieutenant (operator or API).
+
+
+==== New Release of a Commodore Component
+
+The new release of a Commodore component won't trigger anything directly.
```
corvus-ch, line 139: Do we have already explicitly stated that commodore component will get proper releases and we no longer will just pull latest master? If not, it might be worth be a bit more explicit here. If not part of the discussions, one might not get what is meant by this paragraph.

docs/modules/SDDs/pages/0021-cluster-catalog-compilation.adoc
```diff
@@ -0,0 +1,170 @@
...
+To manually trigger a compile for developping or debugging purposes the webhook can be called with a tool like `curl`.
+Providing a cluster or tenant ID can limit the set of clusters to compile.
+
+
+=== Risks and Mitigations
+
+==== Rate Limiting
+
+Since the compilation is a relatively expensive operation resource wise, it might become necessary to implement some kind of rate limiting.
+For example could Quartermaster make sure that a certain amount of compile jobs per amount of time isn't exceeded for a single cluster or tenant.
```
corvus-ch:
```suggestion
For example Quartermaster could make sure that a certain amount of compile jobs per amount of time isn't exceeded for a single cluster or tenant.
```

corvus-ch (changes requested):

docs/modules/SDDs/pages/0021-cluster-catalog-compilation.adoc
srueg: But then we don't know if the job ran or not.
Also we just move the storing of information from the API to the clusters, which probably makes it harder to do.

corvus-ch: Is it really the responsibility of Quartermaster to know which job ran and which not? This is something monitoring is for. Besides that, how would that true/false flag help Quatermaster to do its decision? What information are you talking about that would need to be stored on the cluster?

docs/modules/SDDs/pages/0021-cluster-catalog-compilation.adoc
srueg: GitHub as well as GitLab do `POST` requests by default for webhooks.

docs/modules/SDDs/pages/0021-cluster-catalog-compilation.adoc
srueg: I added the numbers because it's the order in which they happen.

docs/modules/SDDs/pages/0021-cluster-catalog-compilation.adoc
corvus-ch: But still, you can not configure the `Authorization` header. Github allows you do define a _secret_ which is then submitted in the header  `X-Hub-Signature`.

I phrase this as a more general question: how do we want authentication to work?

docs/modules/SDDs/pages/0021-cluster-catalog-compilation.adoc
srueg: Let's do this in the components docs / how-to write components

docs/modules/SDDs/pages/0021-cluster-catalog-compilation.adoc
srueg: I added a note to make sure the logic when/how the job is run is the same on clusters and the central system.
How exactly this will look like can be defined at a later point.

docs/modules/SDDs/pages/0021-cluster-catalog-compilation.adoc
srueg: Outcome of the discussion:
We need to know for which "compile request" the compilation was run/needs to run. So neither a boolean nor a timestamp are the right data types I guess.

We could use a unique ID (i.e. compile-ID) which changes for each compile request. That way the Quartermaster can just bump the ID if a new compile needs to run and the system that runs the jobs knows if for a certain request the compile was already run.

corvus-ch: Works for me. Just a side note: in this context, a unique ID and a timestamp are equivalent.

srueg: Almost: a unique ID can be made unique, while a timestamp might collide (if created at the same time).

corvus-ch: Please update the text. Right now, it still specifies that we use a boolean and we agreed to not do so.

corvus-ch (approved):

PR #52, opened 2020-06-16
Improve getting started guide

tobru:


akosma:
I would add a small paragraph explaining what a "tenant" is; it's the only concept that seems a bit obscure after re-reading this text.

PR #55, opened 2020-06-19
SDD 0022 Managed Services in Cluster

tobru:
This SDD adds a design concept to allow managed services in a Kubernetes cluster.

### DIFF
--- a/docs/modules/SDDs/pages/0022-managed-services-in-cluster.adoc
+++ b/docs/modules/SDDs/pages/0022-managed-services-in-cluster.adoc
@@ -1,0 +1,175 @@
+= SDD 0022 - Managed Services in Cluster
+
+:sdd_author:    Tobias Brunner
+:sdd_owner:
+:sdd_reviewers:
+:sdd_date:      2020-06-19
+:sdd_status:    draft
+include::partial$meta-info-table.adoc[]
+
+[NOTE]
+.Summary
+====
+Besides the system services Project Syn also provides services to the user of the platform. This SDD describes how some kind of managed services on Kubernetes can be achieved with Project Syn. It's considered to be the initial version of this concept.
+====
+
+== Motivation
+
+While the current state of Project Syn focuses on delivering system services to Kubernetes clusters, it should also take application services into consideration. There needs to be a way to install, configure and maintain application services and make them available to the user of the platform.
+
+As application services are considered services like databases, caches, queues and other similar applications (sometimes called "Middleware"). They're usually used by the end user application running on a Kubernetes cluster, but could also be used by applications running outside of the cluster (for example in another cluster or even a VM).
+
+=== Goals
+
+* Offer application services to applications running in or outside of the cluster
+* Maintenance processes to maintain these application services, making them a managed service
+* Definition of how the end user application instantiates and consumes an application service
+* Integration into backup and restore processes
+* Automatic monitoring and alerting of service instances
+
+=== Non-Goals
+
+* All plumbing and system services aren't discussed in this SDD
+* Application instance object definition (custom resource) object management
+
+== Design Proposal
+
+It's assumed that all application services are managed by a specialized Kubernetes Operator, which brings the domain knowledge about the specific service. Each application service Kubernetes Operator will be installed and configured by a Commodore Component. The Commodore Component will be responsible for:
+
+* Installing the Kubernetes Operator
+* Pre-configuring the Kubernetes Operator with best practices, if applicable
+* Deploy additional objects like monitoring definition, backup definition or any other object needed to make the application service a managed service
+* Deploy objects for the managed services controller, discussed in xref:0023-managed-services-controller.adoc[SDD 0023 - Managed Services Controller]
+
+Once the Kubernetes Operator is installed and operational, the user of the platform requests an instance of the application service by creating an Operator specific Kubernetes object (Custom Resource). The job of the Kubernetes Operator is to instantiate the service and provide information to the user for actually connecting to the service.
+
+This SDD doesn't define how this application service instance object is handled and managed. It could come together with the application deployment objects of the customer but could also be stored in a separate GitOps repository, just for these objects.
+
+A Kubernetes Operator could also be responsible to manage different kinds of application services, it's not strictly necessary that each Operator is only responsible for exactly one service (examples are KubeDB or Crossplane).
+
+Each provided service will be documented in a dedicated documentation repo for all services.
+
+=== Terminology
+
+Application Service:: A third party service used and accessed by the main application. This can be a database, cache, queue or the like (sometimes also calle Middleware).
+
+Application Service Instance:: An instance of an application service, like a database instance, represented by a Kubernetes Object in the target Kubernetes Cluster.
+
+Kubernetes Operator:: A piece of software which is tightly integrated into Kubernetes, acting on `CustomResourceDefinition` Kubernetes Object.
+
+=== User Stories
+
+==== Kafka with Strimzi Operator - Application in Cluster
+
+_As the user of the platform I want to use Apache Kafka for my application running in the same Kubernetes cluster as a managed service._
+
+* Project Syn Commodore Component `component-kafka-strimzi` deploys Strimzi to the cluster and configures it according to best practices.
+* The user of the platform creates a `Kafka` object and puts in parameters as needed. The object is stored at the same place as the other application deployment objects and deployed to the cluster with the same mechanisms (for example GitOps).
+* The Strimzi Kubernetes Operator deploys Kafka and creates a secret with the connection parameters for the users application to consume.
+* Via the xref:0023-managed-services-controller.adoc[Managed Services Controller] configuration objects deployed by the Commodore Component the application service is automatically monitored and backed up.
+* For keeping the Strimzi Kubernetes Operator up-to-date, Project Syn mechanisms are used. See xref:0017-maintenance-with-renovate.adoc[SDD 0017 - Maintenance with Renovate] for details.
+* For keeping Kafka up-to-date the user of the platform has to maintain the `Kafka` object.
+
+==== Kafka with Strimzi Operator - Application outside Cluster
+
+_As an application owner I want to use Apache Kafka from my application running outside of the cluster._
+
+Similar to first use case, but:
+
+* The `Kafka` object is stored in _a_ GitOps repository and deployed with the Project Syn provided Argo CD.
+* The connection parameters are handed over to the application owner.
+* To access Kafka the Kubernetes services are configured in a way so that Kafka is reachable from outside of the cluster.
+
+=== Implementation Details/Notes/Constraints
+
+==== Kubernetes Operator Implementation
+
+It's possible that multiple Kubernetes Operator implementations exists for an application service. It's perfectly possible that the user could choose between several Kubernetes Operators for the same application service - depending on the customers needs.
+
+==== Storage location of Instance Definitions - Application-Service-Only Cluster
+
+In the case of having a Kubernetes Cluster which only provides an application service (without running the users application), a central GitOps repository can be used which stores all application instance objects (Kubernetes Custom Resource objects for the Kubernetes Operator). The Argo CD app will be configured to only apply the subdirectory which applies to this Cluster.
+
+As the user of the application service doesn't get access to the cluster, there is no way of tampering with the Argo CD application definition configuration (which would allow to instantiate different instances).
+
+An alternative could be to create a GitOps repository per customer which contains all the service instance objects and allow the customer to do some kind of self-service. This GitOps repository is only used for storing the Custom Resources to instantiate application service instances. A mechanisms must exist that only allowed objects are stored in this GitOps repository when the customer has access to it.
+
+==== Storage location of Instance Definitions - Customer Application Cluster
+
+When the application service runs on the same cluster as the customer application, it's preferred that the Custom Resource object is stored alongside the application deployment definitions. If this isn't the case, the same storage location can be applied as described above.
+
+==== GitOps Integration
+
+A separate Argo CD project will be created for deploying the application services, this way there is a grouping and clear boundary available.
+
+==== Backup/Restore Integration
+
+Each application service needs its own customized backup and restore process. The default is to use https://k8up.io/[K8up] which is usually provided by the default Project Syn tooling.
+
+The K8up objects are usually needed per service instance.
+
+When the application specific Kubernetes Operator already brings native Backup/Restore functionality which can't be easily integrated into K8up, it's to be considered to use this one.
+
+==== Monitoring and Alerting Integration
+
+NOTE: This section is still subject to change, depending on how the main monitoring concept will actually look like.
+
+Each application service should bring its own `ServiceMonitor`, `PodMonitor` and `PrometheusRule` definitions for the Prometheus Operator. These objects are distributed by the Commodore Component which manages the Kubernetes Operator. 
+
+If there are any Grafana Dashboards available, they should also be distributed by the Commodore Component.
+
+==== Enabling of Managed Service
+
+For each application service instance some supporting objects are needed, as described in the sections above. This can be objects for backup, monitoring or others. As it's not always possible to control the creation of such objects, for example when the service instance object is delivered by the customers CI/CD process or GitOps repository, a tool is needed which is able to create them. This tool is described in xref:0023-managed-services-controller.adoc[SDD 0023 - Managed Services Controller].
+
+==== Enforcing Configuration with Admission Controllers
+
+To be able to enforce configuration best practices and policies on instance definition manifests the https://www.openpolicyagent.org/docs/latest/kubernetes-introduction/[Open Policy Agent Gatekeeper] is used which provides validating and mutating webhooks. The Commodore Component which manages the application service Operator can deploy Rego policies which are then picked up by Open Policy Agent.
+
+Example of policies which could be enforced:
+
+* Supported version in `Kafka` to only allow version `>=2.5` and `<3.0`: `spec.kafka.version`.
+* Enforce minimum number of replicas in `Kafka`: `spec.kafka.replicas`
+
+=== Risks and Mitigations
+
+==== Operator Updates and Upgrades
+
+There is a risk when upgrading a Kubernetes Operator that the service instance could get restarted, redeployed, deleted, destroyed, reconfigured, misconfigured or anything else you could imagine as the Kubernetes Operator is the tool which is responsible for taking care of the application service.
+
+To mitigate the risks posed by Kubernetes Operator upgrades, a test procedure which ensures that these upgrades are safe is required. This also includes figuring out if the managed object definitions have changed and whether objects created for the old version are compatible with the new object definition.
+
+It could also be that objects need to be converted before a Kubernetes Operator can be upgraded. This situation would require additional tooling which orchestrates the conversion.
+
+==== Version mismatch
+
+As the service instance object is managed in a different way and not in the reach of the Commodore Component for the Kubernetes Operator, it could happen that the instance object requests a service instance version which isn't supported by the Kubernetes Operator. Additionally, service instance objects may request versions (or options) which aren't supported anymore after a Kubernetes Operator upgrade.
+
+Therefore, it's important to keep track of the service instances and their versions and keep them in sync and comparison to the Kubernetes Operator supported service instance version (for example Strimzi operator version _a_ only supports Kafka versions _x_ and _y_). This is something which might be done via the Lieutenant API and it's inventory capabilities.
+
+== Drawbacks
+
+=== Service instantiation
+
+With this approach, Project Syn currently only handles the Kubernetes Operator itself including some plumbing, but it doesn't handle the actual service instantiation.
+
+== Alternatives
+
+=== Deployment using Helm
+
+Application service instances could be deployed by using Helm charts, instead of using Kubernetes Operators. This way many advanced functionality and domain knowledge would be missing and would have to be implemented otherwise as deployed Helm Charts are static. That means that no active component takes care of the deployed service (for example initialization or reconciliation) and also day-2 operations tasks like upgrades would have to be carried out manually.
+
+=== Deployment using Commodore Components
+
+Commodore Components could be used to generate application service instances by leveraging Jsonnet functionality, not using the Kubernetes Operator. This would make the whole process much more dependent on external tools and also not tightly integrated into the cloud native way of deploying applications. This is comparable to the alternative described above.
+
+=== Other Kubernetes Operators
+
+Some Operators do exist which are also able to generate objects based on conditions:
+
+* https://github.com/vshn/espejo[Espejo]: This operator is able to keep a set of objects available and up-to-date over the whole cluster. It isn't able to generate objects based on conditions or other objects.
+* https://github.com/redhat-cop/resource-locker-operator[Resource Locker Operator]: The main goal of resource locker is to keep objects in place or patch existing objects. It isn't possible to define objects or patches based on conditions or other objects.
+
+== References
+
+* https://strimzi.io[Strimzi Kubernetes Operator]
--- a/docs/modules/SDDs/pages/0023-managed-services-controller.adoc
+++ b/docs/modules/SDDs/pages/0023-managed-services-controller.adoc
@@ -1,0 +1,129 @@
+= SDD 0023 - Managed Services Controller
+
+:sdd_author:    Tobias Brunner
+:sdd_owner:     
+:sdd_reviewers: 
+:sdd_date:      2020-06-22
+:sdd_status:    draft
+include::partial$meta-info-table.adoc[]
+
+[NOTE]
+.Summary
+====
+To automatically manage application services indepent of where the service instance definition is coming from a managed services controller is introduced.
+====
+
+== Motivation
+
+For automatically generating supporting Kubernetes objects to objects which define application service instances - as defined in xref:0022-managed-services-in-cluster.adoc[0022 - Managed Services in Cluster] - a new controller is introduced. This controller watches for known Kubernetes objects and generates supporting Kubernetes objects like backup schedules or monitoring configuration.
+
+=== Goals
+
+* Controller to enable managed services on a Kubernetes cluster
+
+=== Non-Goals
+
+* Define the content of the supporting objects
+
+== Design Proposal
+
+A Kubernetes controller watches for a defined set of Kubernetes objects and acts when they're created, updated or deleted. The controller acts as a templating engine, meaning it can generate a list of Kubernetes object and fill in needed information in placeholders. The templates and the definition which objects to watch are defined as custom resources which configure the controller.
+
+[source,yaml,title=Example]
+--
+apiVersion: syn.tools/v1alpha1
+kind: ConditionalObjects
+metadata:
+  name: KafkaStrimziManagedServiceProd
+spec:
+  applicationInstanceDefinitionObject:
+    apiVersion: kafka.strimzi.io/v1beta1
+    kind: Kafka
+  managedObjectSelector:
+    *metav1.LabelSelector
+  managedObjectNamespaceSelector:
+    *metav1.LabelSelector
+  generateObjects:
+  - apiVersion: monitoring.coreos.com/v1
+    kind: PrometheusRule
+    metadata:
+      name: KafkaMonitoringRules
+    spec:
+      [...]
+  - apiVersion: monitoring.coreos.com/v1
+    kind: ServiceMonitor
+    metadata:
+      name: KafkaServiceMonitor
+    spec:
+      [...]
+--
+
+These `ConditionalObjects` objects are delivered by the Commodore Component which installs the corresponding application instance operator. In this case the Commodore Component `component-kafka-strimzi` would deploy this matching object.
+
+=== User Stories
+
+==== User instantiates an application service instance
+
+A user of the Project Syn enabled Kubernetes cluster creates a `Kafka` object and delivers it to the cluster. The controller is configured to watch these `Kafka` objects and generate a list of supporting objects to enable this service to be a managed service.
+
+=== Implementation Details/Notes/Constraints
+
+==== Opt-Out
+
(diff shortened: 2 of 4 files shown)


docs/modules/SDDs/pages/0022-managed-services-in-cluster.adoc
```diff
@@ -0,0 +1,111 @@
...
+* Definition of how the end user application instantiates and consumes an application service
+
+=== Non-Goals
+
+* Services offered by the cloud are not part of this SDD
+* All plumbing and system services are discussed in other SDDs
+
+== Design Proposal
+
+It is considered that all application services are managed by a specialized Kubernetes Operator, which brings the domain knowledge about the specific service. This Kubernetes Operator will be installed and configured by a Commodore Component. The Commodore Component will be responsible for:
```
corvus-ch:
```suggestion
It is assumed that all application services are managed by a specialized Kubernetes Operator, which brings the domain knowledge about the specific service. This Kubernetes Operator will be installed and configured by a Commodore Component. The Commodore Component will be responsible for:
```

docs/modules/SDDs/pages/0022-managed-services-in-cluster.adoc
```diff
@@ -0,0 +1,111 @@
...
+=== User Stories
+
+==== Kafka with Strimzi Operator - Application in Cluster
+
+_As the user of the platform I want to use Apache Kafka for my application runnign in the same Kubernetes cluster as a managed service._
+
+* Project Syn Commodore Component `component-kafka-strimzi` deploys Strimzi to the cluster and configures it according to best practices.
+* The user of the platform creates a `Kafka` object and puts in parameters as needed. The object is stored at the same place as the other application deployment objects and also deployed to the cluster with the same mechanisms (f.e. GitOps).
+* The Strimzi Kubernetes Operator deploys Kafka and creates a secret with the connection parameters for the users application to consume.
+* Via the monitoring and backup configuration deployed by the Commodore Component the application service is automatically monitored and backed up.
```
corvus-ch: Please elaborate by make examples for those monitoring and backup configuration. Are we talking about Prometheus service and pod monitors and K8up objects? If this is the case, that will not work. At least not by my current understanding of what a component can and can not do.

docs/modules/SDDs/pages/0022-managed-services-in-cluster.adoc
```diff
@@ -0,0 +1,111 @@
...
+
+* The `Kafka` object is stored in _a_ GitOps repository and deployed with the Project Syn provided Argo CD.
+* The connection parameters are handed over to the application owner.
+* To access Kafka the Kubernetes services are configured in a way so that Kafka is reachable from outside of the cluster.
+
+=== Implementation Details/Notes/Constraints
+
+* It is possible that multiple Kubernetes Operator implementations exists for an application service. It is perfectly possible that the user could choose between several Kubernetes Operators for the same application service - depending on the customers needs.
+* In the case of having a Kubernetes Cluster which only provides an application service, a central GitOps repository can be used which stores all application instance objects. As the user of the application service doesn't get access to the cluster, there is no way of tampering with the Argo CD configuration.
+** An alternative could be to create a GitOps repository per customer which contains all the service instance objects and allow the customer to do some kind of self-service.
```
corvus-ch: A repository per customer which then again contains everything? I might misread that and it might be worth to improve on the wording.

docs/modules/SDDs/pages/0022-managed-services-in-cluster.adoc
```diff
@@ -0,0 +1,111 @@
...
+
+=== Service instantiation
+
+With this approach Project Syn currently only handles the Kubernetes Operator itself including some plumbing, but it doesn't handle the actual service instantiation.
+
+== Alternatives
+
+=== Crossplane
+
+One of the cornerstone of Project Syn will be Crossplane to provide fully integrated service provisioning. At the current state Crossplane will mainly be used for provisioning services in the cloud. For provisioning of services directly in the cluster there is still more work to do. Also Crossplane is currently not production ready yet and would need a lot of work to get it integrated. For this reason directly using an application service specific Operator it is considered good enough until Crossplane has matured enough. The end-goal will be to use Crossplane for all-things service provisioning, be it in the cloud or in the cluster.
```
corvus-ch: I do not consider Crossplane to be an alternative. As far as the concept discussed within this SDD, Crossplane is just an operator and all of the above does apply.

docs/modules/SDDs/pages/0022-managed-services-in-cluster.adoc
```diff
@@ -0,0 +1,111 @@
...
+
+As application services are considered services like databases, caches, queues and other similar applications. They are usually used by the end user application running on a Kubernetes cluster, but could also be used by applications running outside of the cluster (f.e. in another cluster or even a VM).
+
+This doesn't include services offered by the cloud.
+
+=== Goals
+
+* Offer application services to applications running in or outside of the cluster
+* Maintenance processes to maintain these application services, making them a managed service
+* Definition of how the end user application instantiates and consumes an application service
```
srueg, line 26: Shouldn't monitoring/alerting also be a goal of this SDD?

docs/modules/SDDs/pages/0022-managed-services-in-cluster.adoc
```diff
@@ -0,0 +1,111 @@
...
+
+[NOTE]
+.Summary
+====
+Besides the system services Project Syn also provides additional services to the user of the platform. This SDD describes how some kind of managed services on Kubernetes with Project Syn can be achieved. It is considered to be the initial version of this concept.
+====
+
+== Motivation
+
+While the current state of Project Syn focuses on delivering system services to Kubernetes clusters, it should also take application services into consideration. There needs to be a way to install, configure and maintain application services and make them available to the user of the platform.
```
corvus-ch, line 18: The term _application service_ is used throughout the document. For me, this is not a well defined thing and needs definition within this document.

docs/modules/SDDs/pages/0022-managed-services-in-cluster.adoc
srueg: And maybe backup/restore?

docs/modules/SDDs/pages/0022-managed-services-in-cluster.adoc
```diff
@@ -0,0 +1,111 @@
...
+
+==== Kafka with Strimzi Operator - Application in Cluster
+
+_As the user of the platform I want to use Apache Kafka for my application runnign in the same Kubernetes cluster as a managed service._
+
+* Project Syn Commodore Component `component-kafka-strimzi` deploys Strimzi to the cluster and configures it according to best practices.
+* The user of the platform creates a `Kafka` object and puts in parameters as needed. The object is stored at the same place as the other application deployment objects and also deployed to the cluster with the same mechanisms (f.e. GitOps).
+* The Strimzi Kubernetes Operator deploys Kafka and creates a secret with the connection parameters for the users application to consume.
+* Via the monitoring and backup configuration deployed by the Commodore Component the application service is automatically monitored and backed up.
+* For keeping the Strimzi Kubernetes Operator up-to-date the Project Syn mechanisms are used.
```
srueg: Cross link to https://github.com/projectsyn/documentation/blob/master/docs/modules/SDDs/pages/0017-maintenance-with-renovate.adoc ?

docs/modules/SDDs/pages/0022-managed-services-in-cluster.adoc
srueg: If the objects are required per CR, then a component on its own couldn't do it, true.
But if a single object is enough, to make sure the operator metrics are scraped which in turn include metrics for all instantiated CRs then it would be possible.
For the first scenario, a dedicated controller (e.g. syntroller) would be necessary.

docs/modules/SDDs/pages/0022-managed-services-in-cluster.adoc
tobru: Yes, I'll add that...

docs/modules/SDDs/pages/0022-managed-services-in-cluster.adoc
tobru: I was missing a better name and therefore used it. I could also imagine to call this "middleware", but this sounds a bit too enterprisey to me. Do you have any other idea? Of course a definition / glossary would also help.

docs/modules/SDDs/pages/0022-managed-services-in-cluster.adoc
corvus-ch: My assumption goes more along the line of: operators has to also provide the objects for monitoring and backup. This is a noble request. In case of Prometheus monitoring, this might mostly be the case. For k8up, not so much. So we need a concept on how to proceed in those cases.

docs/modules/SDDs/pages/0022-managed-services-in-cluster.adoc
corvus-ch: I do not much care about the term used as long as it is well defined. If a general definition is not given, as it is the case for _application service_, one must be made within the scope of the discussion.

I do not have any suggestion if and how to change that term. I assume that once a definition is written down, we might come up with something more suitable.

tobru: I'll add a Terminology section then.

docs/modules/SDDs/pages/0022-managed-services-in-cluster.adoc
tobru: Yeah, right. I'll replace that by other alternatives.

docs/modules/SDDs/pages/0022-managed-services-in-cluster.adoc
tobru: Will improve that.

docs/modules/SDDs/pages/0022-managed-services-in-cluster.adoc
tobru: I've added a specific section about these topics. Let's continue the review / discussion these new sections.

docs/modules/SDDs/pages/0022-managed-services-in-cluster.adoc
```diff
@@ -0,0 +1,158 @@
...
+
+== Design Proposal
+
+It's assumed that all application services are managed by a specialized Kubernetes Operator, which brings the domain knowledge about the specific service. This Kubernetes Operator will be installed and configured by a Commodore Component. The Commodore Component will be responsible for:
+
+* Installing the Kubernetes Operator
+* Pre-configuring the Kubernetes Operator with best practices, if applicable
+* Deploy additional objects like monitoring definition, backup definition or any other object needed to make the application service a managed service
+
+When the Kubernetes Operator is installed and operational, the user of the platform requests an instance of the application service by creating an Operator specific Kubernetes object. The job of the Kubernetes Operator is to instantiate the service and provide the information to the user for actually connecting to the service.
```
srueg:
```suggestion
Once the Kubernetes Operator is installed and operational, the user of the platform requests an instance of the application service by creating an Operator specific Kubernetes object (CR). The job of the Kubernetes Operator is to instantiate the service and provide the information to the user for actually connecting to the service.
```

docs/modules/SDDs/pages/0022-managed-services-in-cluster.adoc
```diff
@@ -0,0 +1,158 @@
...
+
+This SDD doesn't define how this service instance object is handled and managed. It could come together with the application deployment objects of the customer but also being stored in a separate GitOps repository just for these objects.
+
+A Kubernets Operator could also be responsible to manage different kinds of application services, it'sn't strictly necessary that each Operator is only responsible for one service.
```
srueg:
```suggestion
A Kubernets Operator could also be responsible to manage different kinds of application services, it's not strictly necessary that each Operator is only responsible for one service.
```

docs/modules/SDDs/pages/0022-managed-services-in-cluster.adoc
```diff
@@ -0,0 +1,158 @@
...
+:sdd_owner:
+:sdd_reviewers:
+:sdd_date:      2020-06-19
+:sdd_status:    draft
+include::partial$meta-info-table.adoc[]
+
+[NOTE]
+.Summary
+====
+Besides the system services Project Syn also provides additional services to the user of the platform. This SDD describes how some kind of managed services on Kubernetes with Project Syn can be achieved. It's considered to be the initial version of this concept.
```
corvus-ch:
```suggestion
Besides the system services Project Syn also provides services to the user of the platform. This SDD describes how some kind of managed services on Kubernetes can be achieved with Project Syn. It's considered to be the initial version of this concept.
```

docs/modules/SDDs/pages/0022-managed-services-in-cluster.adoc
```diff
@@ -0,0 +1,158 @@
...
+* Pre-configuring the Kubernetes Operator with best practices, if applicable
+* Deploy additional objects like monitoring definition, backup definition or any other object needed to make the application service a managed service
+
+When the Kubernetes Operator is installed and operational, the user of the platform requests an instance of the application service by creating an Operator specific Kubernetes object. The job of the Kubernetes Operator is to instantiate the service and provide the information to the user for actually connecting to the service.
+
+This SDD doesn't define how this service instance object is handled and managed. It could come together with the application deployment objects of the customer but also being stored in a separate GitOps repository just for these objects.
+
+A Kubernets Operator could also be responsible to manage different kinds of application services, it'sn't strictly necessary that each Operator is only responsible for one service.
+
+Each provided service will be documented in the Commodore Component which installs the Kubernetes Operator for the service.
```
srueg: Not sure if that's the right place to document a service?
I think such a documentation would outgrow a component in terms of details. It's also very possible that multiple components are required for a single service.
```suggestion
Each provided service will be documented in a dedicated documentation repo for all services.
```

docs/modules/SDDs/pages/0022-managed-services-in-cluster.adoc
```diff
@@ -0,0 +1,158 @@
...
+Besides the system services Project Syn also provides additional services to the user of the platform. This SDD describes how some kind of managed services on Kubernetes with Project Syn can be achieved. It's considered to be the initial version of this concept.
+====
+
+== Motivation
+
+While the current state of Project Syn focuses on delivering system services to Kubernetes clusters, it should also take application services into consideration. There needs to be a way to install, configure and maintain application services and make them available to the user of the platform.
+
+As application services are considered services like databases, caches, queues and other similar applications. The're usually used by the end user application running on a Kubernetes cluster, but could also be used by applications running outside of the cluster (f.e. in another cluster or even a VM).
+
+This doesn't include services offered by the cloud.
```
corvus-ch: Having Crossplane in mind, how does this differ and therefore has to be excluded?

* We install the operator (e.g. Kafka Strimzi or Crossplane)
* User creates CRs to instantiate a service instance (Kafa, Cache provided by AWS Redis).

I am a ware of the fact that Crossplane still needs to mature. This might also be true for a whole bunch of other operators.

docs/modules/SDDs/pages/0022-managed-services-in-cluster.adoc
```diff
@@ -0,0 +1,158 @@
...
+* The connection parameters are handed over to the application owner.
+* To access Kafka the Kubernetes services are configured in a way so that Kafka is reachable from outside of the cluster.
+
+=== Implementation Details/Notes/Constraints
+
+==== Kubernetes Operator Implementation
+
+It's possible that multiple Kubernetes Operator implementations exists for an application service. It's perfectly possible that the user could choose between several Kubernetes Operators for the same application service - depending on the customers needs.
+
+==== Storage location of Instance Definitions - Application-Service-Only Cluster
```
corvus-ch, line 89: Do we have an Idea of the customers expectation? What do they prefer to do?

* Click in a GUI and done?
* Have full control integrated into their own automation setup?
* Something in between?

So far, this SDD lies the foundation to cover all of the above. I just wonder on what we have to focus.

docs/modules/SDDs/pages/0022-managed-services-in-cluster.adoc
```diff
@@ -0,0 +1,158 @@
...
+
+Each application service needs it's own customized backup and restore process. The default is to use https://k8up.io/[K8up] which is usually provided by the default Project Syn tooling.
+
+The K8up objects are usually needed per service instance.
+
+NOTE: We still need to figure out how these K8up objects are generated. Do we need a special controller for that?
+
+When the application specific Kubernetes Operator already brings native Backup/Restore functionality which can't be easily integrated into K8up, it's to be considered to use this one.
+
+==== Monitoring and Alerting Integration
```
corvus-ch, line 113: How to handle multi tenancy?

docs/modules/SDDs/pages/0022-managed-services-in-cluster.adoc
```diff
@@ -0,0 +1,158 @@
...
+
+With this approach Project Syn currently only handles the Kubernetes Operator itself including some plumbing, but it doesn't handle the actual service instantiation.
+
+== Alternatives
+
+=== Deployment using Helm
+
+Application service instances could be deployed by using Helm charts, instead of using Kubernetes Operators. This way many advanced functionality and domain knowledge would be missing and would have to be implemented otherwise as deployed Helm Charts are static. That means that no active component takes care of the deployed service (f.e. initialization or reconciliation) and also day-2 operations tasks like upgrades would have to be carried out manually.
+
+=== Deployment using Commodore Components
```
corvus-ch, line 162: I see this as just one way on how customers can instantiate their services.

* Do it manually (discouraged by why not)
* Using any automation/GitOps setup the customer has in place and is familiar with
* Use of Project Syn (customer managed)
* Use of Project Syn (VSHN managed)

How about writing a Jsonnet library? We then can reuse them as we need (render within a commodore component, render OpenShift templates, Helm charts etc.)

docs/modules/SDDs/pages/0022-managed-services-in-cluster.adoc
tobru: You're right. After giving this another thought there is no reason to exclude cloud services. I'll update / remove the corresponding parts.

docs/modules/SDDs/pages/0022-managed-services-in-cluster.adoc
tobru: For the time being it's no GUI, it's just Kubernetes objects coming from different places. I'm not sure if this needs to be reflected in the SDD, what do you think?

docs/modules/SDDs/pages/0022-managed-services-in-cluster.adoc
tobru: Do you think this should be part of this SDD? IMHO this is part of the overall monitoring concept.

docs/modules/SDDs/pages/0022-managed-services-in-cluster.adoc
tobru: Not sure what I should do with this comment. This section talks about alternatives. What do you feel isn't clear or should be updated?

docs/modules/SDDs/pages/0022-managed-services-in-cluster.adoc
corvus-ch: I am not asking for a GUI, just for the direction this is likely to evolve. This is not strictly necessary but could be of value.

docs/modules/SDDs/pages/0022-managed-services-in-cluster.adoc
corvus-ch: Yes, indeed. We really need to tackle that subject soon.

corvus-ch: Do we need to add a reference to the monitoring concept?

docs/modules/SDDs/pages/0022-managed-services-in-cluster.adoc
corvus-ch: Probably I was just musing head. This might become relevant when we implement his.

docs/modules/SDDs/pages/0022-managed-services-in-cluster.adoc
tobru: Yes, I'll add that once we have it. There specifically is a hint on this section, stating that it needs refinement.

docs/modules/SDDs/pages/0022-managed-services-in-cluster.adoc
tobru: I'll leave it out of this SDD, as I think we would have to tackle this in another SDD.

docs/modules/SDDs/pages/0022-managed-services-in-cluster.adoc
```diff
@@ -0,0 +1,166 @@
...
+
+As the user of the application service doesn't get access to the cluster, there is no way of tampering with the Argo CD application definition configuration (which would allow to instantiate different instances).
+
+An alternative could be to create a GitOps repository per customer which contains all the service instance objects and allow the customer to do some kind of self-service. This GitOps repository is only used for storing the Custom Resources to instantiate application service instances. A mechanisms must exist that only allowed objects are stored in this GitOps repository when the customer has access to it.
+
+==== Storage location of Instance Definitions - Customer Application Cluster
+
+When the application service runs on the same cluster as the customer application, it's preferred that the Custom Resource object is stored besides the application deployment definitions. If this isn't the case, the same storage location can be applied as described above.
```
simu:
```suggestion
When the application service runs on the same cluster as the customer application, it's preferred that the Custom Resource object is stored alongside the application deployment definitions. If this isn't the case, the same storage location can be applied as described above.
```

docs/modules/SDDs/pages/0022-managed-services-in-cluster.adoc
```diff
@@ -0,0 +1,166 @@
...
+
+==== GitOps Integration
+
+A separate Argo CD project will be created for deploying the application services, this way there is a grouping and clear boundary available.
+
+==== Backup/Restore Integration
+
+Each application service needs it's own customized backup and restore process. The default is to use https://k8up.io/[K8up] which is usually provided by the default Project Syn tooling.
```
simu:
```suggestion
Each application service needs its own customized backup and restore process. The default is to use https://k8up.io/[K8up] which is usually provided by the default Project Syn tooling.
```

docs/modules/SDDs/pages/0022-managed-services-in-cluster.adoc
```diff
@@ -0,0 +1,166 @@
...
+
+The K8up objects are usually needed per service instance.
+
+When the application specific Kubernetes Operator already brings native Backup/Restore functionality which can't be easily integrated into K8up, it's to be considered to use this one.
+
+==== Monitoring and Alerting Integration
+
+NOTE: This section is still subject to change, depending on how the main monitoring concept will actually look like.
+
+Each application service should bring it's own `ServiceMonitor`, `PodMonitor` and `PrometheusRule` definitions for the Prometheus Operator. These objects are distributed by the Commodore Component which manages the Kubernetes Operator.
```
simu:
```suggestion
Each application service should bring its own `ServiceMonitor`, `PodMonitor` and `PrometheusRule` definitions for the Prometheus Operator. These objects are distributed by the Commodore Component which manages the Kubernetes Operator. 
```

docs/modules/SDDs/pages/0022-managed-services-in-cluster.adoc
```diff
@@ -0,0 +1,166 @@
...
+
+==== Enabling of Managed Service
+
+For each application service instance some supporting objects are needed, as described in the sections above. This can be objects for backup, monitoring or others. As it's not always possible to control the creation of such objects, f.e. when the service instance object is delivered by the customers CI/CD process or GitOps repository, a tool is needed which is able to create them. This tool is described in xref:0023-managed-services-controller.adoc[SDD 0023 - Managed Services Controller].
+
+=== Risks and Mitigations
+
+==== Operator Updates and Upgrades
+
+There is a risk when upgrading a Kubernetes Operator that the service instance could get restarted, redeployed, deleted, destroyed, reconfigured, misconfigured or anything else you could imagine as the Kubernetes Operator is the responsible tool taking care of the application service.
```
simu:
```suggestion
There is a risk when upgrading a Kubernetes Operator that the service instance could get restarted, redeployed, deleted, destroyed, reconfigured, misconfigured or anything else you could imagine as the Kubernetes Operator is the tool which is responsible for taking care of the application service.
```

docs/modules/SDDs/pages/0022-managed-services-in-cluster.adoc
```diff
@@ -0,0 +1,166 @@
...
+
+To mitigate this a defined test procedure is needed which makes sure that Kubernetes Operator upgrades are safe. This also includes figuring out if objects have changed and are still compatible.
```
simu:
```suggestion
To mitigate the risks posed by Kubernetes Operator upgrdes, a test procedure which ensures that these upgrades are safe is required. This also includes figuring out if the managed object definitions have changed and whether objects created for the old version are compatible with the new object definition.
```

docs/modules/SDDs/pages/0022-managed-services-in-cluster.adoc
```diff
@@ -0,0 +1,166 @@
...
+
+It could also be that object conversion are needed before a Kubernetes Operator can be upgraded which would need additional tooling.
```
simu:
```suggestion
It could also be that objects need to be converted before a Kubernetes Operator can be upgraded. This situation would require additional tooling which orchestrates the conversion.
```

docs/modules/SDDs/pages/0022-managed-services-in-cluster.adoc
```diff
@@ -0,0 +1,166 @@
...
+
+==== Version mismatch
+
+As the service instance object is managed in a different way and not in the reach of the Commodore Component for the Kubernetes Operator, it could happen that the instance object requests a version which isn't supported by the Kubernetes Operator or that after an upgrade the version isn't support anymore.
```
simu:
```suggestion
As the service instance object is managed in a different way and not in the reach of the Commodore Component for the Kubernetes Operator, it could happen that the instance object requests a version which isn't supported by the Kubernetes Operator. Additionally, service instance objects may request versions (or options) which are not supported anymore after a Kubernetes Operator upgrade.
```

docs/modules/SDDs/pages/0022-managed-services-in-cluster.adoc
```diff
@@ -0,0 +1,166 @@
...
+
+It's therefore important to keep track of the service instances and their versions and keep them in sync and comparison to the Kubernetes Operator version. This is something which might be done via the Lieutenant API and it's inventory.
```
simu:
```suggestion
Therefore, it's important to keep track of the service instances and their versions and keep them in sync and comparison to the Kubernetes Operator version. This is something which might be done via the Lieutenant API and it's inventory.
```

simu: What do you mean by "keep them in comparison to the Kubernetes Operator version"?

docs/modules/SDDs/pages/0023-managed-services-controller.adoc
```diff
@@ -0,0 +1,89 @@
...
+
+* Controller to enable managed services on a Kubernetes cluster
+
+=== Non-Goals
+
+* Define the content of the supporting objects
+
+== Design Proposal
+
+A Kubernetes controller watches for a defined set of Kubernetes objects and acts when they're created, updated or deleted. The controller acts as a templating engine, meaning it can generate a list of Kubernetes object and fill in needed information in placeholders. The templates and the definition which objects to watch is defined as custom resources which configures the controller.
```
simu:
```suggestion
A Kubernetes controller watches for a defined set of Kubernetes objects and acts when they're created, updated or deleted. The controller acts as a templating engine, meaning it can generate a list of Kubernetes object and fill in needed information in placeholders. The templates and the definition which objects to watch are defined as custom resources which configure the controller.
```

docs/modules/SDDs/pages/0023-managed-services-controller.adoc
```diff
@@ -0,0 +1,89 @@
...
+
+With the annotation `syn.io/managed=false` on a service instance object an opt-out of the managed services controller is possible. No supporting objects will be created if this annotation is set. Default is to be always enabled.
+
+==== Controller CRDs
+
+TODO
+
+=== Risks and Mitigations
+
+We're inventing yet-another templating engine. But as there hasn't been a controller around which implements the requested feature, it has to be done this way.
```
simu:
```suggestion
We're inventing yet another templating engine. However, there is no established controller which implements the features required, it seems unavoidable at this time to implement our own.
```

docs/modules/SDDs/pages/0022-managed-services-in-cluster.adoc
```diff
@@ -0,0 +1,166 @@
...
+.Summary
+====
+Besides the system services Project Syn also provides services to the user of the platform. This SDD describes how some kind of managed services on Kubernetes can be achieved with Project Syn. It's considered to be the initial version of this concept.
+====
+
+== Motivation
+
+While the current state of Project Syn focuses on delivering system services to Kubernetes clusters, it should also take application services into consideration. There needs to be a way to install, configure and maintain application services and make them available to the user of the platform.
+
+As application services are considered services like databases, caches, queues and other similar applications (sometimes called "Middleware"). They're usually used by the end user application running on a Kubernetes cluster, but could also be used by applications running outside of the cluster (f.e. in another cluster or even a VM).
```
simu: "f.e." is not a thing in English. We should use the established "e.g." when shortening "for example"

simu (approved):
LGTM. Mostly English nitpicking and some rewording suggestions.

docs/modules/SDDs/pages/0022-managed-services-in-cluster.adoc
tobru: Added some more text to make this more clear.

corvus-ch (approved):

srueg (approved):

PR #67, opened 2020-06-29
Mass accept SDDs

tobru:
All these SDDs have been discussed in the PRs and are now in state accepted.

### DIFF
--- a/docs/modules/SDDs/pages/0006-commodore-configuration-catalog-compilation.adoc
+++ b/docs/modules/SDDs/pages/0006-commodore-configuration-catalog-compilation.adoc
@@ -4,7 +4,7 @@
 :sdd_owner:     Simon Gerber
 :sdd_reviewers:
 :sdd_date:      2019-10-18
-:sdd_status:    accepted
+:sdd_status:    implemented
 include::partial$meta-info-table.adoc[]
 
 [NOTE]
--- a/docs/modules/SDDs/pages/0006-commodore-configuration-catalog-compilation.adoc
+++ b/docs/modules/SDDs/pages/0006-commodore-configuration-catalog-compilation.adoc
@@ -4,7 +4,7 @@
 :sdd_owner:     Simon Gerber
 :sdd_reviewers:
 :sdd_date:      2019-10-18
-:sdd_status:    draft
+:sdd_status:    accepted
 include::partial$meta-info-table.adoc[]
 
 [NOTE]
--- a/docs/modules/SDDs/pages/0007-lieutenant-management-api.adoc
+++ b/docs/modules/SDDs/pages/0007-lieutenant-management-api.adoc
@@ -4,7 +4,7 @@
 :sdd_owner:     SIG Syn
 :sdd_reviewers:
 :sdd_date:      2019-10-18
-:sdd_status:    draft
+:sdd_status:    accepted
 include::partial$meta-info-table.adoc[]
 
 [NOTE]
--- a/docs/modules/SDDs/pages/0007-lieutenant-management-api.adoc
+++ b/docs/modules/SDDs/pages/0007-lieutenant-management-api.adoc
@@ -4,7 +4,7 @@
 :sdd_owner:     SIG Syn
 :sdd_reviewers:
 :sdd_date:      2019-10-18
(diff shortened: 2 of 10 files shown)


srueg (approved):

corvus-ch (approved):

docs/modules/SDDs/pages/0006-commodore-configuration-catalog-compilation.adoc
```diff
@@ -4,7 +4,7 @@
 :sdd_owner:     Simon Gerber
 :sdd_reviewers:
 :sdd_date:      2019-10-18
-:sdd_status:    draft
+:sdd_status:    accepted
```
simu: In my opinion, Commodore implements everything described in this SDD. Shouldn't we mark the SDD as "implemented" in that case?

The [SDD Definition](https://syn.tools/syn/SDDs/0001-sdd-definition.html) is not clear about the SDD procedure when changes are made to a tool described by an SDD in state "accepted" or "implemented". What's the process when making changes to a tool or process which implements (parts of) an accepted/implemented SDD?

docs/modules/SDDs/pages/0007-lieutenant-management-api.adoc
```diff
@@ -4,7 +4,7 @@
 :sdd_owner:     SIG Syn
 :sdd_reviewers:
 :sdd_date:      2019-10-18
-:sdd_status:    draft
+:sdd_status:    accepted
```
simu: What's the status of the Lieutenant implementation compared to the SDD? Can we mark this as "implemented"?

docs/modules/SDDs/pages/0009-steward-cluster-agent.adoc
```diff
@@ -4,7 +4,7 @@
 :sdd_owner:     Simon Rüegg
 :sdd_reviewers: 
 :sdd_date:      2019-10-18
-:sdd_status:    draft
+:sdd_status:    accepted
```
simu: How does the Steward implementation match up with this SDD? Could this SDD be considered to be marked "implemented"?

docs/modules/SDDs/pages/0019-gitops.adoc
```diff
@@ -4,7 +4,7 @@
 :sdd_owner:
 :sdd_reviewers:
 :sdd_date:      2019-11-20
-:sdd_status:    speculative
+:sdd_status:    accepted
```
simu: Does the current ArgoCD setup cover everything here? Should we mark this as "implemented"?

simu (commented):
WDYT about marking some of the SDDs as "implemented"?

At least Commodore matches the goals and design described in SDD 0006 well enough to consider the SDD "implemented" in my opinion.

docs/modules/SDDs/pages/0006-commodore-configuration-catalog-compilation.adoc
tobru: You're right, I completely forgot about the "implemented" state. This certainly makes sense here. IMHO SDDs capture a point-in-time when this was discussed. It isn't the status-quo documentation, it helps for understanding how something was "invented". The actual status-quo should be documented alongside the tool (which isn't really done today). For changing something a new SDD should be written. If something actually doesn't apply anymore or has been replaced by a new concept (SDD!) then an SDD should be marked "obsolete".

docs/modules/SDDs/pages/0019-gitops.adoc
tobru: yes

docs/modules/SDDs/pages/0009-steward-cluster-agent.adoc
tobru: yes

docs/modules/SDDs/pages/0007-lieutenant-management-api.adoc
tobru: yes

tobru:
Updated SDD 0006, 0007, 0009 and 0019 to be marked as "implemented"

simu (approved):

PR #71, opened 2020-07-07
SDD describing the automated billing process

tobru:


tobru:
This has never been worked out in more detail and nowadays, we use the so-called "APPUiO Cloud billing system" which might be named "Syntax" in the future - or not.

PR #73, opened 2020-07-22
Apply the Syn default meta files

corvus-ch:


tobru:
I'm not sure if a CHANGELOG makes sense here and if we should treat this repo 1:1 like the others. What do other think?

corvus-ch:
> I'm not sure if a CHANGELOG makes sense here

I thought about this. If I am not mistaken, we do have a changelog for the VSHN handbook. In my opinion, it can be of value.

srueg:
I guess the changelog only makes sense if we plan to have releases of the docs. Not sure this makes sense for documentation...

corvus-ch:
I have removed the change log. We can add that again when we have a need for it. Please review and approve/request changes.

tobru (approved):

srueg (approved):

PR #100, opened 2020-10-20
docs(getting started): gitlab username is required

mweibel:
got me confused for a second: a gitlab user id is something different from the username - the userid is in profile settings and can't be used to later access the created repository.


mweibel:
FYI failed editorconfig check seems not to fail because of my PR, but because of another existing docs page.

tobru:
Thanks a lot!

> FYI failed editorconfig check seems not to fail because of my PR, but because of another existing docs page.

This is being fixed in #99

PR #101, opened 2020-10-20
Update Getting Started Guide for newer Project Syn tools releases

tobru:
This PR updates the Getting Started guide to work with the latest releases of Project Syn

## Checklist

- [X] Keep pull requests small so they can be easily reviewed.
- [X] Update the documentation.

### DIFF
--- a/docs/modules/ROOT/pages/tutorials/getting-started.adoc
+++ b/docs/modules/ROOT/pages/tutorials/getting-started.adoc
@@ -58,10 +58,10 @@ Install *Lieutenant Operator* with the following commands:
 [source,shell]
 ----
 # CRDs (global scope)
-kubectl apply -k "github.com/projectsyn/lieutenant-operator/deploy/crds?ref=v0.2.0"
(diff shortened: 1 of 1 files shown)


ccremer:
Question: Does it make sense to have a make target or so which runs a script to get the latest releases? So we could set placeholders in the adoc-files and replace with newest values easily. GH actions would then verify if there are newer releases?

Similar to https://github.com/appuio/container-oc/blob/master/src/release.sh where we query the GitHub API for getting latest releases

tobru:
> Question: Does it make sense to have a make target or so which runs a script to get the latest releases? So we could set placeholders in the adoc-files and replace with newest values easily. GH actions would then verify if there are newer releases?

At the current stage of the project just patching the version doesn't really help. As you can see in the WIP commits it's a lot more work than just bumping version numbers. That's not something I'm surprised about, we're still in the 0.x development cycle where compatibility breaks merely all the time. As things settle down we can look in automating this again.

srueg (approved):

simu (approved):

PR #131, opened 2021-04-20
Add how-to page to compile a cluster

ccremer:
* Adds a dedicated how-to page to compile clusters including setting environment variables.

## Checklist

### DIFF
--- a/docs/modules/ROOT/pages/how-tos/compile-catalog.adoc
+++ b/docs/modules/ROOT/pages/how-tos/compile-catalog.adoc
@@ -1,0 +1,76 @@
+= Compile a catalog
+
+Compile and push a cluster catalog from local machine to update dependencies.
+
+== Requirements
+
+* A https://syn.tools/commodore/running-commodore.html[working commodore command].
+* An access token to a running Lieutenant instance. Storing the token as a kubeconfig is recommended.
+* An existing Cluster in Lieutenant.
+
+== Prepare Commodore
+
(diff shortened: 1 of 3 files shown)


docs/modules/ROOT/pages/how-tos/compile-catalog.adoc
```diff
@@ -0,0 +1,76 @@
...
+----
+
+== Compilation
+
+.Choose cluster
+. Get the Cluster ID from Lieutenant.
++
+[source,bash]
+----
+kubectl get cluster -n <lieutenant-instance>
```
simu: Alternatively `commodore catalog list -v`

simu (approved):
LGTM, but I'm not sure this needs to be separate from "Change a parameter".

docs/modules/ROOT/pages/how-tos/compile-catalog.adoc
ccremer: Would the LIEUTENANT_NS variable still be needed?

ccremer:
> LGTM, but I'm not sure this needs to be separate from "Change a parameter".

They're very similar indeed. I could update the change-parameter guide so it builds on top of compile-catalog.

docs/modules/ROOT/pages/how-tos/compile-catalog.adoc
ccremer: I just compiled without LIEUTENANT_NS and it still worked.

PR #135, opened 2021-06-29
align makefile with other projects

tobru:
This PR aligns the Makefile with other projects we're using. It mainly makes sure that `make preview` is available to have the local preview up and running, instead of `make docs-serve`, like in any other Antora documentation project we're having.

## Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,15 +7,10 @@ on:
 jobs:
   linting:
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        command:
-          - lint_yaml
-          - docs-vale
     steps:
       - uses: actions/checkout@v2
-      - name: Run ${{ matrix.command }}
-        run: make ${{ matrix.command }}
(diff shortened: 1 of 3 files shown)


ccremer:
I'm not sure which projects you are referring to, but for the components it's more of a disalignment now :D
(the components use make docs-serve, also in modulesync template f.e.)

tobru:
Oh noes... Then, let's forget about it

ccremer:
not so hasty :)
I'm not opposed to the change. in fact it seems to improve certain areas and we could consider it making into the modulesync and component template so that at least new and m-managed components benefit from new spellers etc.

tobru:
I didn't mean to invest too much time, and I'm unsure if it's feasible for other non-docs-only related projects. As I'm currently mostly working with docs-only repos (aka Makefiles), the most common command I'm using is `make preview`, which on this project would be `make docs-serve`. Having `make preview` on a project where documentation isn't the primary intent might not make sense.

PR #144, opened 2021-10-04
Add explanation page for Commodore component best practices

simu:
## Checklist


docs/modules/ROOT/pages/explanations/commodore-components.adoc
```diff
@@ -0,0 +1,49 @@
...
+    // Transform legacy string parameter to object
+    {
+      name: 'FOO_VALUE', <1>
+      value: params.foo,
+    }
+  else
+    // Assume object type by default and use the provided object
+    params.foo;
+----
+<1> As mentioned above, we inject `name: 'FOO_VALUE'` to preserve the old behavior of the component for configurations which haven't been updated to the new parameter structure yet.
```
ccremer: Is it worth adding another example that _adds_ a new variable?
Example
```yaml
secret: my-password
```
to
```yaml
secretRef:
  name: secret-name-with-password
```
- What would be the precedence if both are defined?
- Should this be treated on a case by case basis or can we recommend a best practice?

docs/modules/ROOT/pages/explanations/commodore-components.adoc
```diff
@@ -0,0 +1,49 @@
+= Commodore component best practices
+
+== Handling major changes
```
ccremer: In a separate PR or commit, I would also welcome a page about handling Helm charts, or more in general how to better support different Kubernetes distributions.
I'm thinking along the "lookup table" where we configure different YAML blocks based on a single switch.

docs/modules/ROOT/pages/explanations/commodore-components.adoc
```diff
@@ -0,0 +1,49 @@
...
+
+=== Structural changes of a component's existing parameters
+
+It's considered best practice to support both the old and new structure of existing parameters, when a structural change is required.
+Structural change can be detected in Jsonnet by inspecting the parameters type or present fields.
+
+To identify a parameter's type in Jsonnet the standard library function https://jsonnet.org/ref/stdlib.html#type[`std.type()`], or the convenience wrappers, such as `std.isString()` can be used.
+To inspect the fields present in a parameter of type `object`, the Jsonnet function `std.objectFields()` can be used.
+
+The parameter's type can then be used to transform the old structure into the new structure in the implementation, giving users a window to migrate their configurations to the new structure.
```
corvus-ch: Possibly drop the key words _forward only migration_ so one can do a read up on the subject and get a better understanding of the idea.

docs/modules/ROOT/pages/explanations/commodore-components.adoc
corvus-ch: I would say we should recommend a best practice. However I am uncertain how practical it is.

The new structure should take precedence unless it comes from the component defaults. 🤔

docs/modules/ROOT/pages/explanations/commodore-components.adoc
ccremer: Another suggestion for a new section: Handling secrets like we do now with https://github.com/appuio/component-openshift4-authentication/pull/32

simu: added in latest version

docs/modules/ROOT/pages/explanations/commodore-components.adoc
simu: added in latest version, but not right here.

docs/modules/ROOT/pages/explanations/commodore-components.adoc
simu: Added a section discussing best practices for replacing parameters.

docs/modules/ROOT/pages/explanations/commodore-components.adoc
simu: I'll add another section in the best practices guide in an additional commit. Or is the page already getting to the point where we should we split it up into subpages?

ccremer: Subpages makes sense I think. In the navigation, we can group them together under best practices, e.g. 
```
Explanation
  Commodore Components Best Practice
    * Secrets
    * Major changes
    ...
```

docs/modules/ROOT/pages/explanations/commodore-components.adoc
```diff
@@ -0,0 +1,3 @@
+= Commodore component best practices
+
+This section of the documentation lists best practices to consider when implementing Commodore components.
```
ccremer: I don't think we need this page at all

docs/modules/ROOT/nav.adoc
```diff
@@ -88,6 +88,12 @@ include::steward:ROOT:partial$nav-reference.adoc[]
 
 .Explanation
 * xref:SDDs:index.adoc[Syn Design Documents]
+* xref:explanations/commodore-components.adoc[Commodore Component Best Practices]
```
ccremer: no link to a placeholder page necessary.
```suggestion
* Commodore Component Best Practices
```

docs/modules/ROOT/pages/explanations/commodore-components/container-images.adoc
```diff
@@ -0,0 +1,33 @@
+= Managing container image versions
+
+It's considered best practice to provide component parameter `images` which contains the default container image versions for all images used by the component.
+
+NOTE: This structure may become mandatory to support automated component dependency updates using Renovate.
```
ccremer: When I discussed this feature with Simon Rüegg, we thought that this requirement shouldn't be. It should be possible for Renovate to detect any images in the parameter structure, if they follow a certain convention, e.g. the keyword `image` followed by at least `repository` and `tag` maybe.
```yaml
parameters:
  someComponent:
    someOtherKey:
      image:
        registry: docker.io
        repository: company/imagename
        tag: latest
```
The reason to lift this limitation is for example when Helm values are exposed: Renovate can then make PRs to update those as well. Also we'd avoid having to duplicate the image in `parameters.images` and then reference via helm values like `parameters.helm_values.image.repository: ${mycomponent:images:someimage:repository}`

TL;DR: I would remove the note and update this page once we know for sure.

ccremer (approved):

docs/modules/ROOT/pages/explanations/commodore-components/major.adoc
```diff
@@ -0,0 +1,93 @@
+= Handling major changes in Commodore components
+
+Generally, it's considered best practice to avoid changes which require manual intervention when upgrading the component.
+
+This document provides best practices to avoid requiring manual intervention during the component upgrade when making changes of the following types:
+
+. Major version upgrades of dependencies
```
corvus-ch:
```suggestion
. <Major version upgrades of dependencies>
```

Asciidoctor creates anchor links for headers. By wrapping a matching text in `<…>` creates a link to that section.

Alternatively you can use the document attribute `:toc:`. The caveat there is, you can not control its placement. The teable of content will always at the beginning of the document.

docs/modules/ROOT/pages/explanations/commodore-components/major.adoc
```diff
@@ -0,0 +1,93 @@
...
+. Structural changes of the component's existing parameters
```
corvus-ch:
```suggestion
. <Structural changes of the component's existing parameters>
```

docs/modules/ROOT/pages/explanations/commodore-components/major.adoc
```diff
@@ -0,0 +1,93 @@
...
+. Replacing component parameters
```
corvus-ch:
```suggestion
. <Replacing component parameters>
```

corvus-ch (approved):

PR #150, opened 2022-02-22
Update best practice format for Helm chart specifications

simu:
The new format makes the implementation for extracting Helm chart dependencies in commodore-renovate much easier.

## Checklist


glrf (approved):
LGTM.

One thing I'm unsure about is the not very explicit requirement that the charts key needs to equal the chart_name for renovate to work. However the alternative of having an explicit `chart_name` field in the `defaults.yml` is also not great (and ironically what we just talked about being a code smell :sweat_smile: ). So I'm fine with leaving it as is.

simu:
> One thing I'm unsure about is the not very explicit requirement that the charts key needs to equal the chart_name for renovate to work. However the alternative of having an explicit `chart_name` field in the `defaults.yml` is also not great (and ironically what we just talked about being a code smell sweat_smile ). So I'm fine with leaving it as is.

Added an extra `IMPORTANT` box noting that the name must match for Renovate to be able to extract the dependency.

PR #155, opened 2022-04-25
Update "Compile a catalog" how-to for OIDC-enabled Lieutenant instances

simu:
We update the how-to to provide instructions for Lieutenant instances
which use OIDC tokens and Lieutenant instances which use long-lived
Kubernetes tokens.

## Checklist

### DIFF
--- a/docs/modules/ROOT/pages/how-tos/compile-catalog.adoc
+++ b/docs/modules/ROOT/pages/how-tos/compile-catalog.adoc
@@ -14,6 +14,26 @@ The following snippet sets up needed environment variables in your current worki
 
 TIP: You can skip this step if you have already defined them already with another method.
 
+. Setup required environment variables
++
+.Lieutenant instance using OIDC token
+[%collapsible]
+====
+[source,bash]
+----
+LIEUTENANT_URL="the-public-lieutenant-API-URL"
+
+cat << EOF > .env
+LIEUTENANT_AUTH="Authorization:Bearer \$(commodore fetch-token)"
+LIEUTENANT_URL="${LIEUTENANT_URL}"
+COMMODORE_API_URL="${LIEUTENANT_URL}"
+EOF
+----
+====
++
+.Lieutenant instance using long-lived Kubernetes token
+[%collapsible]
+====
 [source,bash]
 ----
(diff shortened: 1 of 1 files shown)


docs/modules/ROOT/pages/how-tos/compile-catalog.adoc
```diff
@@ -14,6 +14,27 @@ The following snippet sets up needed environment variables in your current worki
...
 
+. Setup required environment variables
++
+.Lieutenant instance using OIDC token
+[%collapsible]
+====
+[source,bash]
+----
+LIEUTENANT_URL="the-public-lieutenant-API-URL"
+LIEUTENANT_AUTH='Authorization:Bearer $(commodore fetch-token)'
```
glrf: Shouldn't we put this directly in the `.env` file. Otherwise the token will be outdated the second time we try to activate the environment

simu: That's exactly why we need the single-quoted version outside the HERE-doc block. Using

```
cat << EOF > .env
LIEUTENANT_AUTH="Authorization:Bearer $(commodore fetch-token)"
EOF
```

writes a token to the file instead of the command.

simu: Just realized that I can simply escape the `$()` with `\$()` in the HERE-doc. Updated the PR accordingly.

glrf (approved):

docs/modules/ROOT/pages/how-tos/compile-catalog.adoc
simu: Hm, it's even more complicated, having `$(commodore fetch-token)` in the env file will evaluate the command when sourcing the file, which is probably not really what you want when following a long-ish how-to. However, the only way I found to evaluate the command when the variable is used is with `curl -H"$(eval ${LIEUTENANT_AUTH})"`.

Using that approach we'd have to check all our how-tos and update them where necessary.

glrf: IMO having `$(commodore fetch-token)` in the env file is good enough for now.

PR #157, opened 2022-04-29
SDD 0028 - Reusable Commodore Component Configuration Packages

simu:
## Checklist

### DIFF
--- a/docs/modules/SDDs/pages/0028-reusable-config-packages.adoc
+++ b/docs/modules/SDDs/pages/0028-reusable-config-packages.adoc
@@ -40,51 +40,68 @@ While this approach addresses distribution within an organization, it falls shor
 
 == Design Proposal
 
-* Regardless of discovery/fetching/inclusion in the hierarchy packages will be Git repos
+We propose that each Commodore component configuration package is a Git repository.
+This gives us change history and versioning of configuration packages based on a well-established toolset.
+Making each configuration package a Git repository also fits well with the Project Syn approach of keeping all dependencies (global and tenant config, components) in Git.
+We can also profit from all the already existing logic for cutting releases, and CI tests which we've engineered for Commodore components.
 
-Approach 1:
+The big design question is revolves around the mechanism used to fetch configuration packages.
+We can either fetch configuration packages using Kapitan's remote inventory support or we can enhance Commodore's native dependency fetching logic to support config packages in addition to components.
 
-* Use Kaptian's remote inventory support to include package classes directly in the hierarchy, cf. https://kapitan.dev/inventory/#inventory-classes, section "Remote Inventory"
-* (+) We reuse an existing Kapitan feature
-* (+) Only a few additional lines of custom code required
-* (-) Catalog compilation becomes more convoluted (more calls to Kapitan interspersed & we need to explicitly call Kapitan's `fetch_inventories`)
-* (-) We need to live with the package fetching options provided by Kapitan
+=== Approach 1: Use Kapitan's remote inventory support
 
-Approach 2:
+The first option for fetching config packages during catalog compilation is based on Kapitan's remote inventory support.
+By reusing an existing Kapitan feature we get nice integration into the config hierarchy for free.
+Also, by integrating a Kapitan feature, the amount of new code in Commodore is fairly small (on the order of tens of lines).
 
-* Build our own inventory fetching using a similar structure to component fetching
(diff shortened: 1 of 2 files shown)


glrf (approved):
This looks like a good approach.

As for the question of using remote inventories or implementing a custom solution: If I understand the approach correctly, using Kapitan remote inventories will lead to package fetching working differently than a user would expect (e.g. not being able to overwrite already included packages).

I think I would slightly prefer the custom approach. This will lead to a bit more custom code, but probably leads to a better user experience.

docs/modules/SDDs/pages/0028-reusable-config-packages.adoc
```diff
@@ -0,0 +1,127 @@
...
+* Programmatic manipulation of component defaults
+
+== Design Proposal
+
+We propose that each Commodore component configuration package is a Git repository.
+This gives us change history and versioning of configuration packages based on a well-established toolset.
+Making each configuration package a Git repository also fits well with the Project Syn approach of keeping all dependencies (global and tenant config, components) in Git.
+We can also profit from all the already existing logic for cutting releases, and CI tests which we've engineered for Commodore components.
+
+The big design question is revolves around the mechanism used to fetch configuration packages.
```
corvus-ch:
```suggestion
The big design question revolves around the mechanism used to fetch configuration packages.
```

corvus-ch (approved):

simu:
> This looks like a good approach.
> 
> As for the question of using remote inventories or implementing a custom solution: If I understand the approach correctly, using Kapitan remote inventories will lead to package fetching working differently than a user would expect (e.g. not being able to overwrite already included packages).
> 
> I think I would slightly prefer the custom approach. This will lead to a bit more custom code, but probably leads to a better user experience.

Please see the latest version. I've now defined the package specification and inclusion more explicitly. This changes the trade-off between Kapitan's remote inventory and Commodore-native remote inventory a bit, as we now have to transform the user-provided config into Kapitan's format if we want to use Kapitan's remote inventory. My guess is that this will reduce the difference in implementation complexity between the native and Kapitan fetching from the earlier proposal.

docs/modules/SDDs/pages/0028-reusable-config-packages.adoc
```diff
@@ -45,33 +45,49 @@ This gives us change history and versioning of configuration packages based on a
 Making each configuration package a Git repository also fits well with the Project Syn approach of keeping all dependencies (global and tenant config, components) in Git.
 We can also profit from all the already existing logic for cutting releases, and CI tests which we've engineered for Commodore components.
 
+=== Specifying configuration packages
+
+We propose a new top-level parameter `configuration_packages`.
```
glrf: Why not just `packages`?

glrf:
> Please see the latest version. I've now defined the package specification and inclusion more explicitly. This changes the trade-off between Kapitan's remote inventory and Commodore-native remote inventory a bit, as we now have to transform the user-provided config into Kapitan's format if we want to use Kapitan's remote inventory. My guess is that this will reduce the difference in implementation complexity between the native and Kapitan fetching from the earlier proposal

Looks good, this is more or less how I expected the custom approach to work. You're right we can probably make it work the same way when using Kapitan's remote inventory. It just feels a bit "hacky"? I would probably still go with the custom approach, but I don't really have strong feelings either way

docs/modules/SDDs/pages/0028-reusable-config-packages.adoc
simu: I've no hard feelings either way, I'd be fine with using `packages`.

simu: Changed to `packages` in the latest version

simu:
> > Please see the latest version. I've now defined the package specification and inclusion more explicitly. This changes the trade-off between Kapitan's remote inventory and Commodore-native remote inventory a bit, as we now have to transform the user-provided config into Kapitan's format if we want to use Kapitan's remote inventory. My guess is that this will reduce the difference in implementation complexity between the native and Kapitan fetching from the earlier proposal
> 
> Looks good, this is more or less how I expected the custom approach to work. You're right we can probably make it work the same way when using Kapitan's remote inventory. It just feels a bit "hacky"? I would probably still go with the custom approach, but I don't really have strong feelings either way

I'm tending towards the Commodore-native implementation as well, since I think we'll end up with a cleaner implementation, even if there's a bit more of it.

PR #172, opened 2023-02-13
Update Project Syn Roadmap for 2023 and beyond

tobru:


CODEOWNERS
```diff
@@ -1 +1 @@
-* @projectsyn/tarazed
+* @projectsyn/aldebaran
```
simu: Core Project Syn isn't currently owned by Aldebaran, so this change doesn't make sense to me.

simu (commented):
Documentation changes LGTM, the CODEOWNERS change doesn't make sense since the current ownership of Project Syn is with the VSHN Tech Alignment WG and not Team Aldebaran.

CODEOWNERS
tobru: You're right, I've now chosen project-syn-ig.

simu (approved):

PR #181, opened 2024-06-13
Add SDD for Commodore Compile Pipeline

HappyTetrahedron:
## Checklist

### DIFF
--- a/docs/modules/SDDs/pages/0032-compile-pipeline.adoc
+++ b/docs/modules/SDDs/pages/0032-compile-pipeline.adoc
@@ -1,0 +1,153 @@
+= SDD 0031 - Central Component Version tracking
+
+:sdd_author:    Aline Abler
+:sdd_owner:     Project Syn IG
+:sdd_date:      2024-06-13
+:sdd_status:    draft
+
+include::partial$meta-info-table.adoc[]
+
+[NOTE]
+.Summary
+====
+This describes how we want to implement CI/CD for Project Syn using a Commodore compile pipeline, which enables automatic compilation of cluster catalogs whenever the corresponding tenant repository is modified.
+
+Furthermore, it explains how want to extend the functionality of Lieutenant to enable it to automatically configure the compile pipeline on tenant repositories.
+====
+
+== Motivation
+
+Having a continuous integration solution unlocks a number of benefits:
+It solves the problem of configuration drift, where changes to the tenant repository might not be reflected in every cluster catalog because not all of them have since been compiled, and it lessens the burden on catalog maintainers who otherwise would need to locally compile each cluster individually.
+
+It is already fairly straightforward to manually set up basic auto-compilation for individual tenant repositories without special support from Project Syn itself.
+At VSHN, such a solution has been in use for several years.
+
+However, certain features (such as automatic configuration of the compile pipeline) are hard to implement in a standalone fashion.
+As such features are now desired, it makes sense to fully integrate the compile pipeline into Project Syn.
+
+By making Project Syn "CI-aware", we can implement more seamless management of compile pipeline configuration on the tenant repositories, including automated setup and automated token rotation.
+This will go hand-in-hand with the existing repository management features in Lieutenant.
+
+Currently, we're looking at a solution that is specific to GitLab CI/CD, and to tenant and catalog repositories that are stored on GitLab.
+Other CI solutions and other repository hosts might be supported in the future.
+
+=== Goals
+
+* Provide pipeline definitions for GitLab CI/CD to automatically compile and push cluster catalogs on a tenant repository.
+* Enable Lieutenant to autonomously manage the configuration required to set up the compile pipeline for a cluster catalog.
+
+=== Non-Goals
+
+* Support for CI solutions other than GitLab CI/CD.
+
+== Design Proposal
+
+=== Gitlab CI/CD Pipeline for Commodore
+
+The pipeline definitions for the Commodore compile pipeline are managed in a dedicated GitLab repository (the pipeline repository).
+Tenant repositories can then include these pipeline definitions in their `gitlab-ci.yml`.
+
+Execution of the pipeline is controlled via GitLab CI/CD Variables, which can be set on the tenant repository.
+The variables control which catalog is compiled, and are used to provide access to the catalog repository via GitLab Project Access Token, as well as to the Lieutenant API via Commodore API token.
+
+In particular, the pipeline accepts the following CI/CD variables:
+
+* `ACCESS_TOKEN_CLUSTERNAME`, where `CLUSTERNAME` is the name of a specific cluster, with `-` replaced by `_`. This contains a GitLab Project Access Token, which must have read-write access the corresponding cluster's catalog repository.
+* `COMMODORE_API_URL`. This contains the URL at which the Lieutenant API can be accessed.
+* `COMMODORE_API_TOKEN`. This contains an access token for the Lieutenant API.
+* `CLUSTERS`. This contains a space-separated list of cluster IDs which should be compiled and pushed automatically.
+
+=== CRD
+
+We add a new field to the `GitRepoTemplate` (and, by extension, the `GitRepo`) CRD, under the `.spec` key, called `compilePipeline`.
+
+The `compilePipeline` field contains configuration pertaining to the automatic setup of the compile pipeline on the tenant repository.
+It is optional.
+Absence of the field disables automatic setup and management of the compile pipeline.
+
+The `compilePipeline` field contains a dict with the following fields:
+
+* `gitlabCatalogProjectAccessSecretRef`: (for cluster catalog repos) Reference to a secret containing the GitLab Project Access Token that can be used to access this repository. (Managed by Lieutenant)
+* `pipelineFiles`: Dictionary containing file paths as keys, and file contents as values.
+   These files will be committed to the tenant repository when the compile pipeline is configured, if they do not exist already.
+
+We add a new field to the `Cluster` CRD, under the `.status` key, called `compileMeta`.
+
+
+[source,yaml]
+----
+apiVersion: syn.tools/v1alpha1
+kind: Cluster
+metadata:
+  name: my-cluster
+spec:
+  gitRepoTemplate:
+    compilePipeline:
+      gitlabCatalogProjectAccessSecretRef: c-my-cluster-project-access
+      pipelineFiles:
+        .gitlab-ci.yml: |
+          include:
+            - project: syn/commodore-compile-pipeline
+              ref: master
+              file: /.gitlab/commodore-common.yml
+
+----
+
+=== Operator
+
+The Lieutenant Operator will be extended to automatically manage the compile pipeline for repositories where this is enabled (by way of configuring the `compilePipeline` field).
+
+Since the compile pipeline has to interact with both the tenant repo as well as the cluster catalog repo, it must be enabled on both corresponding `gitRepo` definitions for the configuration to be functioning.
+This way, it is possible to enable auto-compilation for some, but not all clusters on a tenant.
+And furthermore, it is possible to easily disable auto-compilation for a whole tenant without having to touch each cluster's `gitRepo` definition.
+
+When `compilePipeline` is configured on a cluster repository (catalog repository), Lieutenant does the following:
+
+* Create a new `Project Access Token` on the catalog repository via the GitLab API, using the cluster repository's configured apiSecretRef.
+* Store the new access token in a new secret.
+* Store a reference to the new secret in `catalogProjectAccessSecretRef`.
+* Add a new CI/CD variable to the corresponding tenant repository via the GitLab API, using the tenant repository's configured apiSecretRef. The value is named `ACCESS_TOKEN_CLUSTERNAME` (where `CLUSTERNAME` is the cluster ID with `-` replaced by `_`) and contains the GitLab project access token as its value.
+* Update the CI/CD variable `CLUSTERS` on the corresponding tenant repository to ensure it includes the cluster's ID.
+
+When `compilePipeline` is configured on a *managed* tenant repository, Lieutenant does the following:
+
+* Render the `pipelineFiles` configured on the tenant repository, and commit them.
+* Add a new CI/CD variable `COMMODORE_API_URL` to the tenant repository, containing the API URL for Lieutenant.
+* Add a new CI/CD variable `COMMODORE_API_TOKEN` to the tenant repository, containing the API token for Lieutenant.
+
+When the `compilePipeline` field is removed from a repository, all the corresponding resources should be cleaned up.
+
+Unmanaged repositories cannot have automatically configured compile pipelines.
+
+==== Rotation of Project Access Tokens
+
+GitLab's project access tokens have a limited lifespan.
+Lieutenant should automatically rotate them as required.
+
+To this end, a scheduled task should run in the operator that checks the token's expiration date via the GitLab API, and rotates any that are close to expiring.
+
+To rotate a token:
+
+* Use the GitLab API to rotate the existing token.
+* Update the corresponding catalog project access secret.
+* Update the corresponding CI/CD variable (`ACCESS_TOKEN_CLUSTERNAME`) in the tenant repository.
+
+=== Implementation Details/Notes/Constraints
+
+Existing compile pipeline configuration::
+If a setup already includes a bunch of tenant repositories with manually configured CI/CD, some care has to be taken to ensure the new implementation can "adopt" this configuration.
++
+In particular, these repositories would already have a working `.gitlab-ci.yml` that probably can be left as-is.
++
+Any existing manually created Project Access Tokens will be superseded by new auto-generated ones.
+This will lead to a bunch of now-unused tokens needing to be cleaned up.
+
+External Catalog Repositories::
+There may be cases where the catalog repositories are not hosted on the same repository host as the tenant repository, in which case API access for the purpose of creating Project Access Tokens is unavailable.
+The Commodore Compile Pipeline can still be used against such catalog repositories by specifying an SSH key to access them.
++
+This can still be configured manually, and the automated configuration would not interfere.
+
+== References
+
--- a/docs/modules/SDDs/pages/0032-compile-pipeline.adoc
+++ b/docs/modules/SDDs/pages/0032-compile-pipeline.adoc
@@ -1,8 +1,8 @@
-= SDD 0031 - Central Component Version tracking
+= SDD 0032 - Commodore Compile Pipeline for GitLab
 
 :sdd_author:    Aline Abler
 :sdd_owner:     Project Syn IG
-:sdd_reviewers: Simon Gerber, Tobias Brunner
+:sdd_reviewers: Simon Gerber, Tobias Brunner, Sebastian Widmer
 :sdd_date:      2024-06-13
 :sdd_status:    draft
 
@@ -11,7 +11,7 @@ include::partial$meta-info-table.adoc[]
 [NOTE]
 .Summary
 ====
-This describes how we want to implement CI/CD for Project Syn using a Commodore compile pipeline, which enables automatic compilation of cluster catalogs whenever the corresponding tenant repository is modified.
+This describes how we want to implement GitLab CI/CD for Project Syn using a Commodore compile pipeline, which enables automatic compilation of cluster catalogs whenever the corresponding tenant repository is modified.
 
 Furthermore, it explains how want to extend the functionality of Lieutenant to enable it to automatically configure the compile pipeline on tenant repositories.
 ====
@@ -174,6 +174,8 @@ The operator will reconcile *Tenants* as follows:
 
 Currently, we're looking at a solution that is specific to GitLab CI/CD, and to tenant and catalog repositories that are stored on GitLab.
 Other CI solutions and other repository hosts might be supported in the future.
+The implementation takes care to gracefully skip the CI steps if not on GitLab.
+We leave the implementation extensible enough to add further Git management tools and CI systems without needing breaking changes.
 
 Existing compile pipeline configuration::
 If a setup already includes a bunch of tenant repositories with manually configured CI/CD, some care has to be taken to ensure the new implementation can "adopt" this configuration.
--- a/docs/modules/SDDs/pages/0032-compile-pipeline.adoc
+++ b/docs/modules/SDDs/pages/0032-compile-pipeline.adoc
@@ -4,7 +4,7 @@
 :sdd_owner:     Project Syn IG
 :sdd_reviewers: Simon Gerber, Tobias Brunner, Sebastian Widmer
 :sdd_date:      2024-06-13
-:sdd_status:    draft
+:sdd_status:    accepted
 
 include::partial$meta-info-table.adoc[]
 
@@ -34,12 +34,13 @@ This will go hand-in-hand with the existing repository management features in Li
 
 === Goals
 
-* Provide pipeline definitions for GitLab CI/CD to automatically compile and push cluster catalogs on a tenant repository.
+* Specify the interface that Commodore CI pipeline definitions need to implement in order to work with the managed CI/CD configuration
 * Enable Lieutenant to autonomously manage the configuration required to set up the compile pipeline for a cluster catalog.
 
 === Non-Goals
 
 * Support for CI solutions other than GitLab CI/CD.
+* Provide pipeline definitions to automatically compile and push cluster catalogs on a tenant repository.
 
 == Design Proposal
 
--- a/docs/modules/SDDs/pages/0032-compile-pipeline.adoc
+++ b/docs/modules/SDDs/pages/0032-compile-pipeline.adoc
@@ -2,7 +2,7 @@
 
 :sdd_author:    Aline Abler
 :sdd_owner:     Project Syn IG
-:sdd_reviewers: Simon Gerber
+:sdd_reviewers: Simon Gerber, Tobias Brunner
 :sdd_date:      2024-06-13
 :sdd_status:    draft
 
@@ -62,8 +62,10 @@ The `accessTokenSecretName` field contains a reference to a secret.
 If it is set, the Lieutenant operator will store an access token into this secret, which can be used to access the Git repository.
 In the case of GitLab, this would be a Project Access Token with read-write access to the repository.
 
-The `ciVariables` field contains a dictionary describing variable names and corresponding values.
-These variables are added to the git repository as CI/CD variables.
+The `ciVariables` field contains a list of objects describing variable names and corresponding values.
+Each object in the list has a type that's modeled after a Kubernetes container's `env` field.
+In contrast to container environment variables, our variables only support specifying values directly (via field `value`) or by referencing a `Secret` resource (via field `valueFrom.secretKeyRef`).
+These variables are added to the Git repository as CI/CD variables.
 
 [source,yaml]
 ----
@@ -113,6 +115,9 @@ The `compilePipeline` field contains a dict with the following fields:
 
 * `clusters`: List of cluster IDs of clusters for which the compile pipeline should be executed.
   This field is managed by the operator.
+* `pipelineFiles`: Dictionary containing file paths as keys, and file contents as values.
+  These files will be added to the tenant's `gitRepoTemplate.templateFiles` by the Lieutenant operator.
+  This field is optional; if absent, no new template files are added to the `gitRepoTemplate`.
 
 [source,yaml]
 ----
@@ -124,35 +129,17 @@ spec:
   compilePipeline:
     clusters:
       - c-my-cluster
-----
-
-=== In-Repo CI/CD pipeline configuration
-
-Configuring the CI pipeline usually happens through files committed to the corresponding repository.
-For a Lieutenant-managed pipeline configuration, these files should be managed by Lieutenant.
-To achieve this, we can leverage the existing mechanism to commit template files to git repositories:
-
-[source,yaml]
-----
-apiVersion: syn.tools/v1alpha1
-kind: Tenant
-metadata:
-  name: t-my-tenant
-spec:
-  gitRepoTemplate:
-    templateFiles:
+    pipelineFiles:
       .gitlab-ci.yml: |
         include:
           - project: syn/commodore-compile-pipeline
             ref: master
             file: /.gitlab/commodore-common.yml
-
 ----
-
 
 === Operator
 
-The Lieutenant Operator will be extended to automatically manage the compile pipeline for repositories where this is enabled (by way of configuring the `compilePipeline` field on the tenant and the `enableCompilePipeline` field on the cluster).
+The Lieutenant Operator will be extended to automatically manage the compile pipeline for repositories where this is enabled (by way of deploying the CI config file in the tenant and the `enableCompilePipeline` field on the cluster).
 
 Since the compile pipeline has to interact with both the tenant repository as well as the cluster catalog repositories, it must be enabled on both corresponding resources for the configuration to be functional.
 This way, it is possible to enable auto-compilation for some, but not all clusters on a tenant.
@@ -162,7 +149,7 @@ The operator will reconcile *GitRepos* as follows:
 * When `spec.accessTokenSecretName` is set, the operator generates an access token for the corresponding repository (via the repository host's API, using the API secret in `.spec.apiSecretRef`), and writes this token into a secret with the given name.
   In the case of GitLab, this is a Project Access Token.
   The operator also runs a scheduled job which refreshes these tokens when they are close to expiring, or when they no longer exist on the repository host.
-* The content of `.spec.ciVariables` is written to the repository's configuration on the git host.
+* The content of `.spec.ciVariables` is written to the repository's configuration on the Git host.
   In the case of GitLab, it is written as CI/CD variables.
 
 NOTE: If the GitRepo is of type `unmanaged`, none of these steps will be executed.
@@ -181,6 +168,7 @@ The operator will reconcile *Tenants* as follows:
 * For each entry in `.spec.compilePipeline.clusters`, another entry is added to the tenant repository GitRepo's `spec.ciVariabes`.
   The key is `ACCESS_TOKEN_CLUSTERNAME`, where `CLUSTERNAME` is the ID of a specific cluster, with `-` replaced by `_`.
   The value is a reference to the secret containing the access token to access that cluster's catalog repository, taken from the secret specified in the catalog GitRepo configuration under `.spec.accessTokenSecretName`.
+* For each entry in `.spec.compilePipeline.pipelineFiles`, a new corresponding entry is added to the tenant's `.spec.gitRepoTemplate.templateFiles`.
 
 === Implementation Details/Notes/Constraints
 
--- a/docs/modules/SDDs/pages/0032-compile-pipeline.adoc
+++ b/docs/modules/SDDs/pages/0032-compile-pipeline.adoc
@@ -1,4 +1,4 @@
-= SDD 0031 - Central Component Version tracking
+= SDD 0032 - Commodore Compile Pipeline
 
 :sdd_author:    Aline Abler
 :sdd_owner:     Project Syn IG
--- a/docs/modules/SDDs/pages/0032-compile-pipeline.adoc
+++ b/docs/modules/SDDs/pages/0032-compile-pipeline.adoc
@@ -30,9 +30,6 @@ As such features are now desired, it makes sense to fully integrate the compile 
 By making Project Syn "CI-aware", we can implement more seamless management of compile pipeline configuration on the tenant repositories, including automated setup and automated token rotation.
 This will go hand-in-hand with the existing repository management features in Lieutenant.
 
-Currently, we're looking at a solution that is specific to GitLab CI/CD, and to tenant and catalog repositories that are stored on GitLab.
-Other CI solutions and other repository hosts might be supported in the future.
-
 === Goals
 
 * Provide pipeline definitions for GitLab CI/CD to automatically compile and push cluster catalogs on a tenant repository.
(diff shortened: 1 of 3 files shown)


docs/modules/SDDs/pages/0032-compile-pipeline.adoc
```diff
@@ -0,0 +1,153 @@
+= SDD 0031 - Central Component Version tracking
+
+:sdd_author:    Aline Abler
+:sdd_owner:     Project Syn IG
```
simu, line 4:
```suggestion
:sdd_owner:     Project Syn IG
:sdd_reviewers: Simon Gerber
```

docs/modules/SDDs/pages/0032-compile-pipeline.adoc
```diff
@@ -0,0 +1,153 @@
...
+At VSHN, such a solution has been in use for several years.
+
+However, certain features (such as automatic configuration of the compile pipeline) are hard to implement in a standalone fashion.
+As such features are now desired, it makes sense to fully integrate the compile pipeline into Project Syn.
+
+By making Project Syn "CI-aware", we can implement more seamless management of compile pipeline configuration on the tenant repositories, including automated setup and automated token rotation.
+This will go hand-in-hand with the existing repository management features in Lieutenant.
+
+Currently, we're looking at a solution that is specific to GitLab CI/CD, and to tenant and catalog repositories that are stored on GitLab.
+Other CI solutions and other repository hosts might be supported in the future.
```
simu: IMO, this paragraph doesn't belong in the motivation. It can go into the "Implementation Details/Notes/Constraints" section.

docs/modules/SDDs/pages/0032-compile-pipeline.adoc
```diff
@@ -0,0 +1,153 @@
...
+
+=== Non-Goals
+
+* Support for CI solutions other than GitLab CI/CD.
+
+== Design Proposal
+
+=== Gitlab CI/CD Pipeline for Commodore
+
+The pipeline definitions for the Commodore compile pipeline are managed in a dedicated GitLab repository (the pipeline repository).
```
simu: Afaict, we don't need to manage the included CI config in a GitLab repo (cf. https://docs.gitlab.com/ee/ci/yaml/includes.html#include-a-single-configuration-file), we can import the config from an arbitrary HTTPS URL.

```suggestion
The pipeline definitions for the Commodore compile pipeline are managed in a dedicated Git repository (the pipeline repository).
This repository can be hosted anywhere.
```

docs/modules/SDDs/pages/0032-compile-pipeline.adoc
```diff
@@ -0,0 +1,153 @@
...
+spec:
+  gitRepoTemplate:
+    compilePipeline:
+      gitlabCatalogProjectAccessSecretRef: c-my-cluster-project-access
+      pipelineFiles:
+        .gitlab-ci.yml: |
+          include:
+            - project: syn/commodore-compile-pipeline
+              ref: master
+              file: /.gitlab/commodore-common.yml
```
simu: I would provide both an example `Tenant` and `Cluster` object here. This example is misleading and doesn't match the explanation of how it's actually supposed to be used in the following section.

docs/modules/SDDs/pages/0032-compile-pipeline.adoc
```diff
@@ -0,0 +1,153 @@
...
+It is optional.
+Absence of the field disables automatic setup and management of the compile pipeline.
+
+The `compilePipeline` field contains a dict with the following fields:
+
+* `gitlabCatalogProjectAccessSecretRef`: (for cluster catalog repos) Reference to a secret containing the GitLab Project Access Token that can be used to access this repository. (Managed by Lieutenant)
+* `pipelineFiles`: Dictionary containing file paths as keys, and file contents as values.
+   These files will be committed to the tenant repository when the compile pipeline is configured, if they do not exist already.
+
+We add a new field to the `Cluster` CRD, under the `.status` key, called `compileMeta`.
```
simu: Copy-pasta?

```suggestion
```

docs/modules/SDDs/pages/0032-compile-pipeline.adoc
```diff
@@ -0,0 +1,153 @@
...
+
+The `compilePipeline` field contains configuration pertaining to the automatic setup of the compile pipeline on the tenant repository.
+It is optional.
+Absence of the field disables automatic setup and management of the compile pipeline.
+
+The `compilePipeline` field contains a dict with the following fields:
+
+* `gitlabCatalogProjectAccessSecretRef`: (for cluster catalog repos) Reference to a secret containing the GitLab Project Access Token that can be used to access this repository. (Managed by Lieutenant)
+* `pipelineFiles`: Dictionary containing file paths as keys, and file contents as values.
+   These files will be committed to the tenant repository when the compile pipeline is configured, if they do not exist already.
```
simu: Please clarify here that both of these fields are optional and trigger certain actions by the operator when they're present.

docs/modules/SDDs/pages/0032-compile-pipeline.adoc
```diff
@@ -0,0 +1,153 @@
...
+              ref: master
+              file: /.gitlab/commodore-common.yml
+
+----
+
+=== Operator
+
+The Lieutenant Operator will be extended to automatically manage the compile pipeline for repositories where this is enabled (by way of configuring the `compilePipeline` field).
+
+Since the compile pipeline has to interact with both the tenant repo as well as the cluster catalog repo, it must be enabled on both corresponding `gitRepo` definitions for the configuration to be functioning.
```
simu: You've used "repo" here, but "repository" elsewhere in the text. I'd normalize to either, but I prefer "repository" for the SDDs.

```suggestion
Since the compile pipeline has to interact with both the tenant repository as well as the cluster catalog repositories, it must be enabled on both corresponding `gitRepo` definitions for the configuration to be functional.
```

docs/modules/SDDs/pages/0032-compile-pipeline.adoc
```diff
@@ -0,0 +1,153 @@
...
+* Store the new access token in a new secret.
+* Store a reference to the new secret in `catalogProjectAccessSecretRef`.
+* Add a new CI/CD variable to the corresponding tenant repository via the GitLab API, using the tenant repository's configured apiSecretRef. The value is named `ACCESS_TOKEN_CLUSTERNAME` (where `CLUSTERNAME` is the cluster ID with `-` replaced by `_`) and contains the GitLab project access token as its value.
+* Update the CI/CD variable `CLUSTERS` on the corresponding tenant repository to ensure it includes the cluster's ID.
+
+When `compilePipeline` is configured on a *managed* tenant repository, Lieutenant does the following:
+
+* Render the `pipelineFiles` configured on the tenant repository, and commit them.
+* Add a new CI/CD variable `COMMODORE_API_URL` to the tenant repository, containing the API URL for Lieutenant.
+* Add a new CI/CD variable `COMMODORE_API_TOKEN` to the tenant repository, containing the API token for Lieutenant.
```
simu:
```suggestion
* Add a new CI/CD variable `COMMODORE_API_TOKEN` to the tenant repository, containing the tenant's API token for Lieutenant.
```

docs/modules/SDDs/pages/0032-compile-pipeline.adoc
```diff
@@ -0,0 +1,153 @@
...
+* Use the GitLab API to rotate the existing token.
+* Update the corresponding catalog project access secret.
+* Update the corresponding CI/CD variable (`ACCESS_TOKEN_CLUSTERNAME`) in the tenant repository.
+
+=== Implementation Details/Notes/Constraints
+
+Existing compile pipeline configuration::
+If a setup already includes a bunch of tenant repositories with manually configured CI/CD, some care has to be taken to ensure the new implementation can "adopt" this configuration.
++
+In particular, these repositories would already have a working `.gitlab-ci.yml` that probably can be left as-is.
```
simu: This may not be the case for our setup with the current proposal, since we currently have the `CLUSTERS` CI variable in the `.gitlab-ci.yml` and not as a CI/CD settings variable. Did you verify which variable value "wins"?

docs/modules/SDDs/pages/0032-compile-pipeline.adoc
```diff
@@ -0,0 +1,153 @@
...
+=== Non-Goals
+
+* Support for CI solutions other than GitLab CI/CD.
+
+== Design Proposal
+
+=== Gitlab CI/CD Pipeline for Commodore
+
+The pipeline definitions for the Commodore compile pipeline are managed in a dedicated GitLab repository (the pipeline repository).
+Tenant repositories can then include these pipeline definitions in their `gitlab-ci.yml`.
```
simu:
```suggestion
Tenant repositories can then include these pipeline definitions in their `gitlab-ci.yml` as a "remote" include.
See the https://docs.gitlab.com/ee/ci/yaml/includes.html#include-a-single-configuration-file[GitLab CI/CD documentation] for details.
```

simu (commented):
## General observations / remarks

* We mention the CI pipeline repo at the beginning, but the rest of the SDD doesn't actually care about that at all, since we just provide a generic "put whatever" field for each tenant repo's CI pipeline configuration.
* Currently it feels like the SDD has two flight-levels: 1) the configuration flight-level which isn't really specific to GitLab (and would most likely work as-is with another Git hosting) and 2) the implementation sketch of how the operator would use the new fields to drive GitLab

## Alternative approach for the additional fields in the CRDs

I think that the proposed approach works decently for the narrow scope of the immediate problem, but it seems to me that it tries to unify configs that don't need to be unified while not being very extensible/composable in regards to possible future additions. Therefore, I'd like to propose an alternative approach for configuring the CI pipeline which may be more composable for future extensions at the cost of more upfront work:

Extend `GitRepo` with:
* `spec.accessTokenSecretName`
* `spec.ciVariables`: a `map[string]string`

Note that extensions to `GitRepo` should automatically be reflected in the `gitRepoTemplate` fields.

Extend `Cluster` and tenant with `spec.compilePipeline`

Contents of `compilePipeline` for `Tenant`:
* `pipelineFiles`: CI configuration to deploy in the tenant repo
* `clusters`: a list of cluster IDs for which the CI pipeline should be configured

Contents of `compilePipeline` for `Cluster`:
* `enable`: Boolean which controls whether the CI pipeline should compile this cluster

With this structure:

* The operator generates an access token when it sees `spec.accessTokenSecretName`on a GitRepo and writes the resulting token into a secret with the given name. The operator runs a scheduled job which checks these secrets and refreshes them when they're close to expiring or don't exist anymore in the GitLab API.
* The operator writes the contents of each `GitRepo`'s `spec.ciVariables` to the GitLab repo as CI/CD variables
* The operator updates the tenant's `spec.compilePipeline.clusters` with the cluster's ID when it sees `compilePipeline.enable=true` on a `Cluster`
* The operator watches the tenant's `compilePipeline` config and
   1. Generates an entry in the tenant's `gitRepoTemplate.templateFiles` for each entry in `pipelineFiles`. Users need to explicitly set the entry in `pipelineFiles` to `{remove}` to delete the file in the repo.
   2. Writes suitable CI variables into the Tenant's `gitRepoTemplate.ciVariables` to configure the CI pipeline for each cluster in `compilePipeline.clusters`

I see a couple advantages here:

* We get better separation of the configs that only make sense for the Tenant's git repo, the configs that make sense for the `Cluster`  and the configs that can be useful for arbitrary Git repos.
* We don't tie ourselves to GitLab in the field naming which leaves us in a better position if we want to extend this to non-GitLab Git hosting in the future.

docs/modules/SDDs/pages/0032-compile-pipeline.adoc
HappyTetrahedron: Yes, the CI/CD variables will win. See also https://docs.gitlab.com/ee/ci/variables/#cicd-variable-precedence

docs/modules/SDDs/pages/0032-compile-pipeline.adoc
HappyTetrahedron: yes 🙃

docs/modules/SDDs/pages/0032-compile-pipeline.adoc
```diff
@@ -0,0 +1,191 @@
...
+
+The `compilePipeline` field contains configuration pertaining to the automatic setup of the compile pipeline on the tenant repository.
+It is optional.
+Absence of the field disables automatic setup and management of the compile pipeline.
+
+The `compilePipeline` field contains a dict with the following fields:
+
+* `clusters`: List of cluster IDs of clusters for which the compile pipeline should be executed.
+  This field is managed by the operator.
+* `pipelineFiles`: Dictionary containing file paths as keys, and file contents as values.
```
HappyTetrahedron: I was wondering if it isn't better to just directly add these to `gitRepoTemplate.templateFiles`, I'm not certain what we gain from this extra step. Thoughts?

simu: I was thinking about that as well when writing the initial comment, I guess if we expect the users to provide the complete contents of each pipeline file anyway, then there's no real need for the extra field.

docs/modules/SDDs/pages/0032-compile-pipeline.adoc
```diff
@@ -0,0 +1,191 @@
...
+
+* Support for CI solutions other than GitLab CI/CD.
+
+== Design Proposal
+
+=== Requirements for Pipeline Configuration
+
+Lieutenant imposes certain assumptions on the configuration of the pipeline:
+Namely, the pipeline has to be set up on the tenant repository by way of adding (arbitrary) files to the repository,
+and it is configured through setting CI/CD variables on the repository.
```
simu:
```suggestion
Namely, the pipeline has to be set up on the tenant repository by way of adding (arbitrary) files to the repository, and it is configured through setting CI/CD variables on the repository.
```

docs/modules/SDDs/pages/0032-compile-pipeline.adoc
```diff
@@ -0,0 +1,191 @@
...
+* `COMMODORE_API_URL`. This contains the URL at which the Lieutenant API can be accessed.
+* `COMMODORE_API_TOKEN`. This contains an access token for the Lieutenant API.
+* `CLUSTERS`. This contains a space-separated list of cluster IDs which should be compiled and pushed automatically.
+
+=== GitRepo CRD
+
+We add two new fields to the `GitRepoTemplate` (and, by extension, the `GitRepo`) CRD, under the `.spec` key, called `accessTokenSecretRef` and `ciVariables`.
+
+The `accessTokenSecretRef` field contains a reference to a secret.
+If it is set, the Lieutenant operator will store an access token into this secret, which can be used to access the git repository.
```
simu:
```suggestion
If it is set, the Lieutenant operator will store an access token into this secret, which can be used to access the Git repository.
```

docs/modules/SDDs/pages/0032-compile-pipeline.adoc
```diff
@@ -0,0 +1,191 @@
...
+In the case of GitLab, this would be a Project Access Token with read-write access to the repository.
+
+The `ciVariables` field contains a dictionary describing variable names and corresponding values.
+These variables are added to the git repository as CI/CD variables.
```
simu: Thinking about this again, it will probably make sense to allow the values to be similar to K8s's `env[*].value` field. That way, we don't need to put secrets verbatim in `spec.ciVariables` but can instead use a K8s secretRef.

docs/modules/SDDs/pages/0032-compile-pipeline.adoc
```diff
@@ -0,0 +1,191 @@
...
+The operator will reconcile *Clusters* as follows:
+
+* When `.spec.enableCompilePipeline` is set to `true`, the tenant's `spec.compilePipeline.clusters` is updated to contain the cluster ID.
+* Similarly, when the field is set to `false` or missing, the tenant's `spec.compilePipeline.clusters` is updated to not contain the cluster ID.
+
+The operator will reconcile *Tenants* as follows:
+
+* When `.spec.compilePipeline` exists and isn't empty, the following entries are added to the tenant repository GitRepo's `.spec.ciVariables`:
+** `COMMODORE_API_URL`, containing the URL at which the Lieutenant API can be accessed.
+** `COMMODORE_API_TOKEN`, containing the tenant's access token for the Lieutenant API.
```
simu:
```suggestion
** `COMMODORE_API_TOKEN`, containing a reference to the secret which contains the tenant's access token for the Lieutenant API.
```

docs/modules/SDDs/pages/0032-compile-pipeline.adoc
```diff
@@ -0,0 +1,191 @@
...
+** `CLUSTERS`, containing a space-separated list of cluster IDs taken directly from `.spec.compilePipeline.clusters`.
+* For each entry in `.spec.compilePipeline.pipelineFiles` whose value isn't `{remove}`, a new corresponding entry is added to the tenant's `.spec.gitRepoTemplate.templateFiles`.
+* For each entry in `.spec.compilePipeline.pipelineFiles` whose value is `{remove}`, the corresponding entry is removed from the tenant's `.spec.gitRepoTemplate.templateFiles`.
```
simu: This doesn't need to be specified here, since we still need to replicate entries whose value is `{remove}` verbatim to `spec.gitRepoTemplate.templateFiles` in order for the file to actually be deleted in the Git repo.

docs/modules/SDDs/pages/0032-compile-pipeline.adoc
```diff
@@ -0,0 +1,191 @@
...
+* For each entry in `.spec.compilePipeline.clusters`, another entry is added to the tenant repository GitRepo's `spec.ciVariabes`.
+  The key is `ACCESS_TOKEN_CLUSTERNAME`, where `CLUSTERNAME` is the ID of a specific cluster, with `-` replaced by `_`.
+  The value is the access token to access that cluster's catalog repository, taken from the secret specified in the catalog GitRepo configuration under `.spec.accessTokenSecretRef`.
```
simu:
```suggestion
  The value is a reference to the secret containing the access token to access that cluster's catalog repository, taken from the secret specified in the catalog GitRepo configuration under `.spec.accessTokenSecretRef`.
```

docs/modules/SDDs/pages/0032-compile-pipeline.adoc
```diff
@@ -0,0 +1,191 @@
...
+  This contains a GitLab Project Access Token, which must have read-write access the corresponding cluster's catalog repository.
+* `COMMODORE_API_URL`. This contains the URL at which the Lieutenant API can be accessed.
+* `COMMODORE_API_TOKEN`. This contains an access token for the Lieutenant API.
+* `CLUSTERS`. This contains a space-separated list of cluster IDs which should be compiled and pushed automatically.
+
+=== GitRepo CRD
+
+We add two new fields to the `GitRepoTemplate` (and, by extension, the `GitRepo`) CRD, under the `.spec` key, called `accessTokenSecretRef` and `ciVariables`.
+
+The `accessTokenSecretRef` field contains a reference to a secret.
```
simu: In my mind, `secretRef` has the connotation of referencing a secret that's managed externally. It may be more obvious that this field contains the name of a secret that's generated by the operator if the field is named `accessTokenSecretName`.

docs/modules/SDDs/pages/0032-compile-pipeline.adoc
HappyTetrahedron: Oh! I misunderstood how it works! All right then

tobru (approved):

docs/modules/SDDs/pages/0032-compile-pipeline.adoc
```diff
@@ -0,0 +1,207 @@
+= SDD 0031 - Central Component Version tracking
+
+:sdd_author:    Aline Abler
+:sdd_owner:     Project Syn IG
+:sdd_reviewers: Simon Gerber
```
simu:
```suggestion
:sdd_reviewers: Simon Gerber, Tobias Brunner
```

docs/modules/SDDs/pages/0032-compile-pipeline.adoc
```diff
@@ -0,0 +1,207 @@
...
+
+=== GitRepo CRD
+
+We add two new fields to the `GitRepoTemplate` (and, by extension, the `GitRepo`) CRD, under the `.spec` key, called `accessTokenSecretName` and `ciVariables`.
+
+The `accessTokenSecretName` field contains a reference to a secret.
+If it is set, the Lieutenant operator will store an access token into this secret, which can be used to access the Git repository.
+In the case of GitLab, this would be a Project Access Token with read-write access to the repository.
+
+The `ciVariables` field contains a dictionary describing variable names and corresponding values.
```
simu: With the latest changes, this sentence doesn't match the example anymore.

```suggestion
The `ciVariables` field contains a list of objects describing variable names and corresponding values.
Each object in the list has a type that's modeled after a Kubernetes container's `env` field.
In contrast to container environment variables, our variables only support specifying values directly (via field `value`) or by referencing a `Secret` resource (via field `valueFrom.secretKeyRef`).
```

Not sure if we should explicitly allow configmaps as source in addition to secrets

docs/modules/SDDs/pages/0032-compile-pipeline.adoc
```diff
@@ -0,0 +1,207 @@
...
+These variables are added to the git repository as CI/CD variables.
```
simu: Afaik, the Git project spells itself with a capital G.

```suggestion
These variables are added to the Git repository as CI/CD variables.
```

docs/modules/SDDs/pages/0032-compile-pipeline.adoc
```diff
@@ -0,0 +1,207 @@
...
+  compilePipeline:
+    clusters:
+      - c-my-cluster
+----
+
+=== In-Repo CI/CD pipeline configuration
+
+Configuring the CI pipeline usually happens through files committed to the corresponding repository.
+For a Lieutenant-managed pipeline configuration, these files should be managed by Lieutenant.
+To achieve this, we can leverage the existing mechanism to commit template files to git repositories:
```
simu:
```suggestion
To achieve this, we can leverage the existing mechanism to commit template files to Git repositories:
```

I've looked at that again, and have discovered that Lieutenant will unconditionally set any entries in `templateFiles` to `{remove}` unless the file name matches a currently existing cluster (`c-existing-cluster-id.yml`) or `common.yml`. Therefore, to make this work we'll need to either adjust Lieutenant to allow "unmanaged" entries in `templateFiles` or we'll need to go back to the initial approach with an additional new field in `spec.compilePipeline` where we can define CI pipeline config files.

See https://github.com/projectsyn/lieutenant-operator/blob/831dba95bb3aa4e6dc83fea70e6ecd0f57fa1bd8/controllers/tenant/git.go#L14-L60 for the current implementation.

docs/modules/SDDs/pages/0032-compile-pipeline.adoc
```diff
@@ -0,0 +1,207 @@
...
+We add a new field to the `Tenant` CRD, under the `.spec` key, called `compilePipeline`.
+
+The `compilePipeline` field contains configuration pertaining to the automatic setup of the compile pipeline on the tenant repository.
+It is optional.
+Absence of the field disables automatic setup and management of the compile pipeline.
+
+The `compilePipeline` field contains a dict with the following fields:
+
+* `clusters`: List of cluster IDs of clusters for which the compile pipeline should be executed.
+  This field is managed by the operator.
```
simu: Would it make sense to move this to `status.compilePipeline` now that it's intended to be fully operator-managed?

docs/modules/SDDs/pages/0032-compile-pipeline.adoc
```diff
@@ -0,0 +1,207 @@
...
+          - project: syn/commodore-compile-pipeline
+            ref: master
+            file: /.gitlab/commodore-common.yml
+
+----
+
+
+=== Operator
+
+The Lieutenant Operator will be extended to automatically manage the compile pipeline for repositories where this is enabled (by way of configuring the `compilePipeline` field on the tenant and the `enableCompilePipeline` field on the cluster).
```
simu:
```suggestion
The Lieutenant Operator will be extended to automatically manage the compile pipeline for repositories where this is enabled (by way of deploying the CI config file in the tenant and the `enableCompilePipeline` field on the cluster).
```

docs/modules/SDDs/pages/0032-compile-pipeline.adoc
```diff
@@ -0,0 +1,207 @@
...
+
+Since the compile pipeline has to interact with both the tenant repository as well as the cluster catalog repositories, it must be enabled on both corresponding resources for the configuration to be functional.
+This way, it is possible to enable auto-compilation for some, but not all clusters on a tenant.
+
+The operator will reconcile *GitRepos* as follows:
+
+* When `spec.accessTokenSecretName` is set, the operator generates an access token for the corresponding repository (via the repository host's API, using the API secret in `.spec.apiSecretRef`), and writes this token into a secret with the given name.
+  In the case of GitLab, this is a Project Access Token.
+  The operator also runs a scheduled job which refreshes these tokens when they are close to expiring, or when they no longer exist on the repository host.
+* The content of `.spec.ciVariables` is written to the repository's configuration on the git host.
```
simu:
```suggestion
* The content of `.spec.ciVariables` is written to the repository's configuration on the Git host.
```

docs/modules/SDDs/pages/0032-compile-pipeline.adoc
HappyTetrahedron: I decided to go back to the old way of having `compilePipeline.pipelineFiles` in the tenant spec.

Small bonus: since the operator will autoremove files that aren't either in the pipeline or have the cluster ID as a name (or are `common.yml`), there will no longer be a need for the user to explicitly set `{remove}`.

docs/modules/SDDs/pages/0032-compile-pipeline.adoc
HappyTetrahedron: Since we're going back to having the `pipelineFiles`, it doesn't make sense to fully move this into status anymore. But we might want to split it into configurable vs. operator managed fields and only put the latter in status. I'm not sure what's better practice - keep them together, or have a strict split between managed and user-defined?

simu: I'm not sure what the best practice is here, maybe @bastjan can chime in?

simu (approved):
LGTM overall now.

docs/modules/SDDs/pages/0032-compile-pipeline.adoc
```diff
@@ -0,0 +1,195 @@
+= SDD 0031 - Central Component Version tracking
+
+:sdd_author:    Aline Abler
+:sdd_owner:     Project Syn IG
+:sdd_reviewers: Simon Gerber, Tobias Brunner
```
bastjan:
```suggestion
:sdd_reviewers: Simon Gerber, Tobias Brunner, Sebastian Widmer
```

docs/modules/SDDs/pages/0032-compile-pipeline.adoc
```diff
@@ -0,0 +1,195 @@
+= SDD 0031 - Central Component Version tracking
```
bastjan:
```suggestion
= SDD 0032 - Commodore Compile Pipeline for GitLab
```

docs/modules/SDDs/pages/index.adoc
```diff
@@ -26,3 +26,4 @@ The all are using the xref:sdd-template.adoc[SDD Template].
 * xref:0028-reusable-config-packages.adoc[0028 - Reusable Commodore Component Configuration Packages]
 * xref:0030-argocd-multitenancy.adoc[0030 - Project Syn ArgoCD Multi-Tenant Support]
 * xref:0031-component-version-tracking.adoc[0031 - Central Component Version tracking]
+* xref:0032-compile-pipeline.adoc[0032 - Commodore Compile Pipeline]
```
bastjan, line 29:
```suggestion
* xref:0032-compile-pipeline.adoc[0032 - Commodore Compile Pipeline for GitLab]
```

docs/modules/SDDs/pages/0032-compile-pipeline.adoc
```diff
@@ -0,0 +1,195 @@
...
+It is optional; not specifying it is equivalent to setting it to `false`.
+
+[source,yaml]
+----
+apiVersion: syn.tools/v1alpha1
+kind: Cluster
+metadata:
+  name: c-my-cluster
+spec:
+  enableCompilePipeline: true
```
bastjan, line 106: Do we want to make it more explicit as in `enableGitLabCompilePipeline` or are we assuming future implementations for GitHub or others also use this flag?

docs/modules/SDDs/pages/0032-compile-pipeline.adoc
bastjan: I prefer fully operator managed fields in `.status` it removes a source of misunderstandings. Not a super firm position since Kubernetes also fully manages a lot of fields in `.spec` in the core manifests.

docs/modules/SDDs/pages/0032-compile-pipeline.adoc
```diff
@@ -0,0 +1,195 @@
...
+  The value is a reference to the secret containing the access token to access that cluster's catalog repository, taken from the secret specified in the catalog GitRepo configuration under `.spec.accessTokenSecretName`.
+* For each entry in `.spec.compilePipeline.pipelineFiles`, a new corresponding entry is added to the tenant's `.spec.gitRepoTemplate.templateFiles`.
+
+=== Implementation Details/Notes/Constraints
+
+Currently, we're looking at a solution that is specific to GitLab CI/CD, and to tenant and catalog repositories that are stored on GitLab.
+Other CI solutions and other repository hosts might be supported in the future.
+
+Existing compile pipeline configuration::
+If a setup already includes a bunch of tenant repositories with manually configured CI/CD, some care has to be taken to ensure the new implementation can "adopt" this configuration.
```
bastjan: I don't necessarily agree that we want/need to adopt anything. I'd rather have consistency and i don't see how having a clean cut would break something.

docs/modules/SDDs/pages/0032-compile-pipeline.adoc
```diff
@@ -0,0 +1,195 @@
...
+** `CLUSTERS`, containing a space-separated list of cluster IDs taken directly from `.spec.compilePipeline.clusters`.
+* For each entry in `.spec.compilePipeline.clusters`, another entry is added to the tenant repository GitRepo's `spec.ciVariabes`.
+  The key is `ACCESS_TOKEN_CLUSTERNAME`, where `CLUSTERNAME` is the ID of a specific cluster, with `-` replaced by `_`.
+  The value is a reference to the secret containing the access token to access that cluster's catalog repository, taken from the secret specified in the catalog GitRepo configuration under `.spec.accessTokenSecretName`.
+* For each entry in `.spec.compilePipeline.pipelineFiles`, a new corresponding entry is added to the tenant's `.spec.gitRepoTemplate.templateFiles`.
+
+=== Implementation Details/Notes/Constraints
+
+Currently, we're looking at a solution that is specific to GitLab CI/CD, and to tenant and catalog repositories that are stored on GitLab.
+Other CI solutions and other repository hosts might be supported in the future.
```
bastjan, line 188:
```suggestion
Other CI solutions and other repository hosts might be supported in the future.
The implementation takes care to gracefully skip the CI steps if not on GitLab.
We leave the implementation extensible enough to add further Git management tools and CI systems without needing breaking changes.
```

docs/modules/SDDs/pages/0032-compile-pipeline.adoc
```diff
@@ -0,0 +1,195 @@
...
+:sdd_reviewers: Simon Gerber, Tobias Brunner
+:sdd_date:      2024-06-13
+:sdd_status:    draft
+
+include::partial$meta-info-table.adoc[]
+
+[NOTE]
+.Summary
+====
+This describes how we want to implement CI/CD for Project Syn using a Commodore compile pipeline, which enables automatic compilation of cluster catalogs whenever the corresponding tenant repository is modified.
```
bastjan, line 14:
```suggestion
This describes how we want to implement GitLab CI/CD for Project Syn using a Commodore compile pipeline, which enables automatic compilation of cluster catalogs whenever the corresponding tenant repository is modified.
```

bastjan (approved):
I fixed and made the title of the SDD more specific for GitLab since there might be a future SDD for GitHub or Forgejo.

docs/modules/SDDs/pages/0032-compile-pipeline.adoc
HappyTetrahedron: From how it's currently designed, I can easily envision future versions using the same flag. We already have a way to distinguish between different git hosts in the gitrepo spec. The rest of the config should be transferable.

docs/modules/SDDs/pages/0032-compile-pipeline.adoc
HappyTetrahedron: Fair enough. At the moment, it's written in a way where adoption will "just work" except perhaps in some exceptional cases where the in-repo config has been manually verbaschtled.
I agree it's not useful to try and account for those latter cases.

simu:
> I fixed and made the title of the SDD more specific for GitLab since there might be a future SDD for GitHub or Forgejo.

If we actually want to make this SDD title specific to GitLab, we'll also need to rename the source file, otherwise it will look very confusing.

Edit: Additionally, I'm not sure this needs to be made specific to GitLab in the title; if there's a future SDD for another Git hosting platform, I'd assume that it would be something along the lines of `SDD 00XY - Forgejo as GitRepo provider` where we can discuss everything that needs to be extended to support this.

Edit 2: Overall, my hope is that the structure we come up with here would be applicable for most Git hosting platforms. Afaict we could easily use the exact same structure discussed here to configure GitHub actions, and iirc Gitea/Forgejo actions are modeled after GH actions so that should just work(tm) as well.

docs/modules/SDDs/pages/0032-compile-pipeline.adoc
HappyTetrahedron: I've rephrased the section a bit to not make it seem like we want to implement extrawurst support for all sorts of existing manual config.

HappyTetrahedron:
IMO adding GitLab to the title has the main benefit of making it immediately clear that other git hosts are not yet supported (even though the architecture is generic enough to enable that in the future).

docs/modules/SDDs/pages/0032-compile-pipeline.adoc
HappyTetrahedron: I've split it up now. This prompted me to also add an `enabled` field in the spec.

simu:
> IMO adding GitLab to the title has the main benefit of making it immediately clear that other git hosts are not yet supported (even though the architecture is generic enough to enable that in the future).

While that's true, explicitly mentioning GitLab in the title also gives the impression that the contents of this SDD aren't applicable for other Git hosting solutions when they clearly are.

HappyTetrahedron:
@simu you have convinced me. I changed it back, but added a sentence at the start specifying that we're only supporting GitLab for now.

bastjan:
I agree that it should be generic enough for other Git/CI hosts. I did find the combination of "Commodore Compile Pipeline" and then out-of-scope "any other systems" and the interweaving of generic and GitLab terms a bit confusing. No hard feelings though not making it more specific to GitLab. (The copy-paste error in the title still needs fixing tho 😉)

docs/modules/SDDs/pages/0032-compile-pipeline.adoc
```diff
@@ -0,0 +1,210 @@
+= SDD 0032 - Commodore Compile Pipeline
+
+:sdd_author:    Aline Abler
+:sdd_owner:     Project Syn IG
+:sdd_reviewers: Simon Gerber, Tobias Brunner, Sebastian Widmer
+:sdd_date:      2024-06-13
+:sdd_status:    draft
```
simu:
```suggestion
:sdd_status:    accepted
```

docs/modules/SDDs/pages/0032-compile-pipeline.adoc
```diff
@@ -0,0 +1,210 @@
...
+
+However, certain features (such as automatic configuration of the compile pipeline) are hard to implement in a standalone fashion.
+As such features are now desired, it makes sense to fully integrate the compile pipeline into Project Syn.
+
+By making Project Syn "CI-aware", we can implement more seamless management of compile pipeline configuration on the tenant repositories, including automated setup and automated token rotation.
+This will go hand-in-hand with the existing repository management features in Lieutenant.
+
+=== Goals
+
+* Provide pipeline definitions for GitLab CI/CD to automatically compile and push cluster catalogs on a tenant repository.
```
simu: This is actually not really achieved anymore with the latest iteration. I'd move this to non-goals for this SDD since we now exclude pipeline definitions completely, and replace it with the following

```suggestion
* Specify the interface that Commodore CI pipeline definitions need to implement in order to work with the managed CI/CD configuration
```

docs/modules/SDDs/pages/0032-compile-pipeline.adoc
```diff
@@ -0,0 +1,210 @@
...
+* Enable Lieutenant to autonomously manage the configuration required to set up the compile pipeline for a cluster catalog.
+
+=== Non-Goals
+
+* Support for CI solutions other than GitLab CI/CD.
```
simu, line 42:
```suggestion
* Support for CI solutions other than GitLab CI/CD.
* Provide pipeline definitions to automatically compile and push cluster catalogs on a tenant repository.
```

simu (approved):

docs/modules/SDDs/pages/0032-compile-pipeline.adoc
HappyTetrahedron: Well spotted 👍🏻

PR #183, opened 2025-01-08
Add SDD for Component Instance Versioning

HappyTetrahedron:
- [x] Link this PR to related issues.

This relates to https://github.com/projectsyn/commodore/issues/563

If you have suggestions for Non-Goals, let me know, I couldn't think of anything.
Also, I'm not sure who I'm supposed to put as Owner.

### DIFF
--- a/docs/modules/SDDs/pages/0033-commodore-component-instance-versioning.adoc
+++ b/docs/modules/SDDs/pages/0033-commodore-component-instance-versioning.adoc
@@ -4,7 +4,7 @@
 :sdd_owner:     Aldebaran
 :sdd_reviewers: Simon Gerber
 :sdd_date:      2025-01-08
-:sdd_status:    draft
+:sdd_status:    accepted
 include::partial$meta-info-table.adoc[]
 
 [NOTE]
--- a/docs/modules/SDDs/pages/0033-commodore-component-instance-versioning.adoc
+++ b/docs/modules/SDDs/pages/0033-commodore-component-instance-versioning.adoc
@@ -1,0 +1,91 @@
+= SDD 0033 - Commodore Component Instance Versioning
+
+:sdd_author:    Aline Abler
+:sdd_owner:     Aline Abler
+:sdd_reviewers: SIG
+:sdd_date:      2025-01-08
+:sdd_status:    draft
+include::partial$meta-info-table.adoc[]
+
+[NOTE]
+.Summary
+====
+This SDD describes possible options to allow specifying different versions for each instance of the same Commodore component.
+It additionally details the options we chose to implement.
+====
+
+== Motivation
+
+With xref:0025-commodore-component-instantiation.adoc[SDD 0025], Commodore can instantiate a component multiple times per cluster if multiple aliases are specified.
+However, all aliases currently use version information from the base component.
+Commodore does not currently support specifying different component versions for each alias.
+
+=== Goals
+
+* Define how different versions for each component instance can be specified and generated.
+
+=== Non-Goals
+
+
+
+== Design Proposal
+
+Commodore will leverage the existing configuration options in `.parameters.components` to retrieve version information for specific instances.
+Instance versions can be specified in `.parameters.components.<instance-name>` and override the value specified in `.parameters.components.<component-name>`.
+Even if all instances provide their own version, a version for the base component (`.parameters.components.<component-name>`) must still be specified.
+
+Component authors must explicitly declare that their component supports multi-versioning.
+In a valid catalog, all used versions of the same component must advertise this support.
+
+Components with multi-versioning support must ensure they create no file conflicts when downloading dependencies.
+In practice, this means components must use the parameter `${_base_directory}` in their compile configuration (`parameters.kapitan.dependencies` and `parameters.kapitan.compile`).
+
+Support for providing different versions of Jsonnet dependencies is currently not implemented.
+Jsonnet dependencies are always provided from the main (non-instantiated) version of the component.
+If a component is to support multi-versioning, Jsonnet dependencies should therefore be compatible within a reasonable span of versions.
+
+Similarly, if the component includes a Jsonnet library that is made available to other components, the version for the library is taken from the main (non-instantiated) version of the component.
+
+=== Implementation Details/Notes/Constraints
+
+We can clone the repository of the base component and create a separate worktree for each instance to check out the correct version.
+
+If the instance version specifies a different URL from the base component, more work is needed: in this case, we clone the changed repository for the instance and create a worktree from that.
+
+This means that if two instances both override the URL to the same value, this repository is cloned twice, which can impact performance.
+However, since this only affects instances which override the repository URL, the impact is tolerable.
+
+=== Risks and Drawbacks
+
+* Since the Jsonnet dependencies and libraries are all provided from a single component version, there is a risk for some combinations of versions not compiling properly due to version compatibility issues.
+** Even if a component's dependencies are sufficiently backwards-compatible at the moment, it's possible for future changes to break compatibility.
+This might force a component author to remove multi-version support in higher component versions, causing a catalog to stop compiling if either component instance is updated.
+* Overriding the repository URL in a component instance always causes a new repository to be cloned - even if the same URL is used in multiple instances.
+
+== Alternatives
+
+=== Automatically detecting multi-version support
+
+It could be possible to automatically determine whether a component supports multi-versioning, by way of checking whether the `${_base_directory}` parameter is used in the component's compile configuration.
+However, automatically declaring components to be "safe for multi-versioning" carries some risk - even if a component is configured accordingly, it is not guaranteed that its provided Jsonnet libraries and/or its Jsonnet dependencies work if used across versions.
+Therefore, we deem it safer to let component authors decide if multi-versioning is safe to use for their component.
+
+=== Supporting different versions of Jsonnet dependencies per instance
+
+We could possibly achieve support for separate Jsonnet dependency versions per instance by reworking how we interface with Kapitan's `compile()` function.
+This would require more extensive reworking of Commodore.
+
+Making this change has no direct impact on the implementation of the multi-versioning feature, so support can still be added at a future time if required.
+
+=== Supporting different versions of a component's provided Jsonnet libraries per instance
+
+It might be possible to allow multiple versions of a component's Jsonnet libraries to be made available to other components.
+Other components would have to refer to the libraries using a specific instance's name.
+This would change how components refer to libraries, and would cause components to be somewhat aware of the context in which they are used in order to refer to the correct instances.
+Since components should be modular and reusable, introducing this sort of coupling goes against the design principles of Project Syn.
+
+== References
+
+* xref:0025-commodore-component-instantiation.adoc[0025 - Commodore Component Instantiation]
+* https://github.com/projectsyn/commodore/issues/563[projectsyn/commodore#563]
+* https://github.com/projectsyn/commodore/pull/559[projectsyn/commodore#559]
--- a/docs/modules/SDDs/pages/0033-commodore-component-instance-versioning.adoc
+++ b/docs/modules/SDDs/pages/0033-commodore-component-instance-versioning.adoc
@@ -1,8 +1,8 @@
 = SDD 0033 - Commodore Component Instance Versioning
 
 :sdd_author:    Aline Abler
-:sdd_owner:     Aline Abler
-:sdd_reviewers: SIG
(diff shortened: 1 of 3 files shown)


docs/modules/SDDs/pages/0033-commodore-component-instance-versioning.adoc
```diff
@@ -0,0 +1,91 @@
...
+This would require more extensive reworking of Commodore.
+
+Making this change has no direct impact on the implementation of the multi-versioning feature, so support can still be added at a future time if required.
+
+=== Supporting different versions of a component's provided JSonnet libraries per instance
+
+It might be possible to allow multiple versions of a component's JSonnet libraries to be made available to other components.
+Other components would have to refer to the libraries using a specific instance's name.
+This would change how components refer to libraries, and would cause components to be somewhat aware of the context in which they are used in order to refer to the correct instances.
+Since components should be modular and reusable, introducing this sort of coupling goes against the design principles of Project Syn.
```
HappyTetrahedron, line 79: Let me know if I'm understanding this correctly. To me this approach just sounded ugly and like a terrible idea.

docs/modules/SDDs/pages/0033-commodore-component-instance-versioning.adoc
```diff
@@ -0,0 +1,91 @@
...
+We can clone the repository of the base component and create a separate worktree for each instance to check out the correct version.
+
+If the instance version specifies a different URL from the base component, more work is needed: in this case, we clone the changed repository for the instance and create a worktree from that.
+
+This means that if two instances both override the URL to the same value, this repository is cloned twice, which can impact performance.
+However, since this only affects instances which override the repository URL, the impact is tolerable.
+
+=== Risks and Drawbacks
+
+* Since the JSonnet dependencies and libraries are all provided from a single component version, there is a risk for some combinations of versions not compiling properly due to version compatibility issues.
```
HappyTetrahedron: I'm not sure off the top of my head how stable these types of dependencies are, so I'm not sure how often we'd expect issues with this.
My current thinking is, we can always implement support for properly versioned JSonnet dependencies when we actually need it - since it looks like quite some effort, and we'll still have the same problem with JSonnet libraries.
Happy to hear your opinions.

docs/modules/SDDs/pages/0033-commodore-component-instance-versioning.adoc
```diff
@@ -0,0 +1,91 @@
...
+** Even if a component's dependencies are sufficiently backwards-compatible at the moment, it's possible for future changes to break compatibility.
+This might force a component author to remove multi-version support in higher component versions, causing a catalog to stop compiling if either component instance is updated.
+* Overriding the repository URL in a component instance always causes a new repository to be cloned - even if the same URL is used in multiple instances.
```
HappyTetrahedron: It's probably possible to optimize this but in my estimation it's more effort than it's worth, since I don't expect this feature to be used super frequently.

docs/modules/SDDs/pages/0033-commodore-component-instance-versioning.adoc
```diff
@@ -0,0 +1,91 @@
...
+
+== Alternatives
+
+=== Automatically detecting multi-version support
+
+It could be possible to automatically determine whether a component supports multi-versioning, by way of checking whether the `${_base_directory}` parameter is used in the component's compile configuration.
+However, automatically declaring components to be "safe for multi-versioning" carries some risk - even if a component is configured accordingly, it is not guaranteed that its provided JSonnet libraries and/or its JSonnet dependencies work if used across versions.
+Therefore, we deem it safer to let component authors decide if multi-versioning is safe to use for their component.
```
HappyTetrahedron: This conclusion is based on my vague understanding of how stable these dependencies typically are, so happy to discuss.

docs/modules/SDDs/pages/0033-commodore-component-instance-versioning.adoc
simu: Please fix JSonnet spelling, should be "Jsonnet" everywhere.

docs/modules/SDDs/pages/0033-commodore-component-instance-versioning.adoc
simu: It should be relatively straightforward to avoid multiple clones of the same repo with the `MultiDependency` infrastructure that's already present in Commodore, but I'd have to look at the code in detail to fully verify this.

Currently we already ensure that we only do a single clone for a monorepo that contains multiple dependencies in different subpaths

docs/modules/SDDs/pages/0033-commodore-component-instance-versioning.adoc
simu: The biggest issue apart from dependency issues -- as far as I can see right now -- is that we'd also have to verify that all operations in the component's Jsonnet use `${_base_directory}` to reference downloaded dependencies (e.g. when using `kap.yaml_load_stream()` to process downloaded alert rules)

docs/modules/SDDs/pages/0033-commodore-component-instance-versioning.adoc
simu: Currently, most components which provide component libraries aren't multi-instance aware (and it rarely makes sense for a component which provides libraries to consumers to be multi-instance aware), so actually supporting different component library versions is mostly an academic consideration.

Nevertheless, the behavior of component libraries in conjunction with multi-versioning capable components needs to be clearly documented, since there's nothing which prevents a multi-instance capable component from providing component libraries.

docs/modules/SDDs/pages/0033-commodore-component-instance-versioning.adoc
simu: Generally speaking, the components which use a lot of Jsonnet dependencies tend to not be multi-instance aware, so the fact that we always install the non-instance version shouldn't be a huge issue for our current components. 

Additionally, we will need to rethink the Jsonnet dependency management at some point anyway (there's other issues with the current approach, such as the fact that different components can't easily use different versions of the same dependency at the moment).

Therefore, while we need to clearly document the behavior of Jsonnet dependencies for component instances with custom versions, I don't think it's currently worth it to spend a lot of time to find a solution which allows instances to use their own Jsonnet dependency versions.

docs/modules/SDDs/pages/0033-commodore-component-instance-versioning.adoc
HappyTetrahedron: Well if it's easy then there's no reason not to do it (and update this doc accordingly).

docs/modules/SDDs/pages/0033-commodore-component-instance-versioning.adoc
```diff
@@ -0,0 +1,91 @@
+= SDD 0033 - Commodore Component Instance Versioning
+
+:sdd_author:    Aline Abler
+:sdd_owner:     Aline Abler
+:sdd_reviewers: SIG
```
simu:
```suggestion
:sdd_reviewers: Simon Gerber
```

docs/modules/SDDs/pages/0033-commodore-component-instance-versioning.adoc
```diff
@@ -0,0 +1,91 @@
+= SDD 0033 - Commodore Component Instance Versioning
+
+:sdd_author:    Aline Abler
+:sdd_owner:     Aline Abler
```
simu: Maybe

```suggestion
:sdd_owner:     Aldebaran
```

docs/modules/SDDs/pages/0033-commodore-component-instance-versioning.adoc
simu: I'd probably remove this bullet point from the design doc.

docs/modules/SDDs/pages/0033-commodore-component-instance-versioning.adoc
```diff
@@ -0,0 +1,91 @@
...
+
+With xref:0025-commodore-component-instantiation.adoc[SDD 0025], Commodore can instantiate a component multiple times per cluster if multiple aliases are specified.
+However, all aliases currently use version information from the base component.
+Commodore does not currently support specifying different component versions for each alias.
+
+=== Goals
+
+* Define how different versions for each component instance can be specified and generated.
+
+=== Non-Goals
```
simu: Remove this section if we have no content for it.

docs/modules/SDDs/pages/0033-commodore-component-instance-versioning.adoc
```diff
@@ -0,0 +1,91 @@
...
+
+
+
+== Design Proposal
+
+Commodore will leverage the existing configuration options in `.parameters.components` to retrieve version information for specific instances.
+Instance versions can be specified in `.parameters.components.<instance-name>` and override the value specified in `.parameters.components.<component-name>`.
+Even if all instances provide their own version, a version for the base component (`.parameters.components.<component-name>`) must still be specified.
+
+Component authors must explicitly declare that their component supports multi-versioning.
```
simu: We should define the `_metadata` field which indicates this support in this document:

```suggestion
Component authors must explicitly declare that their component supports multi-versioning by setting component parameter `_metadata.multi_versioning=true`.
```

docs/modules/SDDs/pages/0033-commodore-component-instance-versioning.adoc
```diff
@@ -0,0 +1,91 @@
...
+Similarly, if the component includes a Jsonnet library that is made available to other components, the version for the library is taken from the main (non-instantiated) version of the component.
+
+=== Implementation Details/Notes/Constraints
+
+We can clone the repository of the base component and create a separate worktree for each instance to check out the correct version.
+
+If the instance version specifies a different URL from the base component, more work is needed: in this case, we clone the changed repository for the instance and create a worktree from that.
+
+This means that if two instances both override the URL to the same value, this repository is cloned twice, which can impact performance.
+However, since this only affects instances which override the repository URL, the impact is tolerable.
```
simu: This paragraph should be updated, as we'll want to implement a variation which only clones each URL once

docs/modules/SDDs/pages/0033-commodore-component-instance-versioning.adoc
```diff
@@ -0,0 +1,91 @@
...
+* Since the Jsonnet dependencies and libraries are all provided from a single component version, there is a risk for some combinations of versions not compiling properly due to version compatibility issues.
+** Even if a component's dependencies are sufficiently backwards-compatible at the moment, it's possible for future changes to break compatibility.
+This might force a component author to remove multi-version support in higher component versions, causing a catalog to stop compiling if either component instance is updated.
+* Overriding the repository URL in a component instance always causes a new repository to be cloned - even if the same URL is used in multiple instances.
+
+== Alternatives
+
+=== Automatically detecting multi-version support
+
+It could be possible to automatically determine whether a component supports multi-versioning, by way of checking whether the `${_base_directory}` parameter is used in the component's compile configuration.
```
simu:
```suggestion
We could try to automatically determine whether a component supports multi-versioning, by way of checking whether the `${_base_directory}` parameter is used in the component's compile configuration and for Jsonnet expressions that refer to downloaded data.
```

docs/modules/SDDs/pages/0033-commodore-component-instance-versioning.adoc
simu:
```suggestion
Additionally, so far we've defined explicit flags for components to indicate that they support optional features, such as multiple instances.
For all of those reasons, we've decided that component authors must explicitly declare whether multi-versioning is safe to use for their component.
```

docs/modules/SDDs/pages/0033-commodore-component-instance-versioning.adoc
```diff
@@ -0,0 +1,85 @@
+= SDD 0033 - Commodore Component Instance Versioning
+
+:sdd_author:    Aline Abler
+:sdd_owner:     Aldebaran
+:sdd_reviewers: Simon Gerber
+:sdd_date:      2025-01-08
+:sdd_status:    draft
```
simu:
```suggestion
:sdd_status:    accepted
```

simu (approved):

docs/modules/SDDs/pages/0033-commodore-component-instance-versioning.adoc
HappyTetrahedron: Ahahaha, very important step!

